PR #60, opened 2021-11-09
Restrict components that are only ever used by the OS

flawedworld:
untested


thestinger:
It's possible that these are meant to be used by other apps.

thestinger:
android.media.action.IMAGE_CAPTURE_SECURE is used by other apps which capture an image via camera on the lockscreen. I don't think it's used by the OS. The ones for opening the camera might also be used beyond double-tap. I don't want to break compatibility and these are officially documented intents so I think we need to expose them.

PR #64, opened 2021-11-17
Style various UI elements to match new GrapheneOS system theme

jorgeluiscarrillo:


thestinger:
We don't control these values outside GrapheneOS and I don't think we can assume they'll work properly.

thestinger:
Ah, I see, you're just hard-wiring them which makes sense.

MHShetty:
Hey @jorgeluiscarrillo,

I have just made the blue a bit brighter as before in our camera application (rest all changes are likely to remain unchanged).

Thanks a lot for contributing to our camera app!

PR #84, opened 2022-01-03
Add French 

Ilithy:
Simple translation in French, I hope not to make a mistake by proposing that.
For any modifications or additions, I remain at your disposal.

### DIFF
--- a/app/src/main/java/app/grapheneos/camera/capturer/ImageCapturer.kt
+++ b/app/src/main/java/app/grapheneos/camera/capturer/ImageCapturer.kt
@@ -101,7 +101,7 @@ class ImageCapturer(private val mActivity: MainActivity) {
             return
         }
 
-        var outputFileOptionsBuilder : ImageCapture.OutputFileOptions.Builder? = null
+        var outputFileOptionsBuilder : ImageCapture.OutputFileOptions.Builder?
 
         try {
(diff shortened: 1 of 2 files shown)


thestinger:
We're going to be making translation teams for the apps and organizing it but we're not able to take translations yet until we have teams for maintaining them. We'll be making a Matrix room for different languages and organizing it from there. We need to have multiple people working on it regularly so strings can be quickly updated when the reference ones in US English are changed.

Ilithy:
@thestinger I couldn't find a thread specific to translations in matrix (I admit I'm lost). If you think that what I have proposed is not good, I will not intervene on any of your projects.

thestinger:
I'm not saying there's any problem with it. Rather, we need to make a translation team for French and other languages which is able to collaborate on translations and review each other's changes. We need to have a group we can ping leading up to a release to finish the translations of new / modified strings.

Ilithy:
I understand, I don't want to spend my days debating this or that translation (for so few elements one or two people are enough), so I let you manage it as you want, good luck and good courage

thestinger:
There will be more strings and changes to them over time, and we need to make sure they're in good shape for all supported languages before doing a release. If we accept translations for a dozen languages, and then change or add 1 string, none of those translations is ready for a release. We need teams we can trust to translate on request leading up to a release.

PR #102, opened 2022-02-01
Add Ukrainian translation

nsherbina1999:


thestinger:
We're going to be making translation teams for the apps and organizing it but we're not able to take translations yet until we have teams for maintaining them. We'll be making a Matrix room for different languages and organizing it from there. We need to have multiple people working on it regularly so strings can be quickly updated when the reference ones in US English are changed.

PR #115, opened 2022-02-14
Update Chinese translation

JohnsonRan:


JohnsonRan:
> We appreciate the contributions, but at the moment we can't take translations yet. We're going to be making translation teams for the apps and organizing it in the future, but we're not able to take translations yet until we have teams for maintaining them. We'll be making a Matrix room for different languages and organizing it from there. We need to have multiple people working on it regularly so strings can be quickly updated when the reference ones in US English are changed.

Please let me know when its ready😃

Zanthed:
We will let you know further information about accepting community translations in the future.

PR #130, opened 2022-03-04
add Chinese translation

abc0922001:


thestinger:
We're interested in this but we need to make a translation team for the languages so that it can be reviewed, maintained and so we can ask for it to be updated after we change the strings. We'll ping you when we're ready to assemble a team for Chinese. We'll try to do this in the near future.

thestinger:
We'll be making Matrix rooms for each translation team, so that would be where this gets coordinated.

Issue #105, opened 2022-02-02
screen allowed to timeout while the camera is on


steve47457:
the device's screen is allowed to dim and turn off (screen timeout) while the camera is on, this should not happen on a camera app. thank you

Samsung Galaxy A03s

thestinger:
There's a full wake lock while recording video. For when the camera preview is enabled, we can add a full wake lock keeping the screen on for a certain amount of time like 5 minutes. I don't think it makes sense to keep the screen / device on indefinitely just because the camera is open with the preview shown.

empratyush:
done #154

PR #141, opened 2022-03-07
Set keepScreenOn to true to prevent screen timeout when camera preview is active

Zanthed:
This covers QR scanning, Camera, and Video modes.

Solves https://github.com/GrapheneOS/Camera/issues/105


anonlatte (approved):

Zanthed:
We want to actually have a longer timeout duration instead of keeping a permanent wake lock for preview like Google Camera does. A user shouldn't accidentally leave this app open and drain their battery.

PR #142, opened 2022-03-07
Remove redundant code, minor cleanups, style fixes

Zanthed:


app/src/main/java/app/grapheneos/camera/ui/activities/MoreSettings.kt
```diff
@@ -122,7 +122,7 @@ class MoreSettings : AppCompatActivity(), TextView.OnEditorActionListener {
 
         snackBar = Snackbar.make(
```
anonlatte, line 123: IMO would be better make variable nullable and init it at `fun showMessage`

app/src/main/java/app/grapheneos/camera/ui/activities/MoreSettings.kt
```diff
@@ -169,44 +168,29 @@ class MoreSettings : AppCompatActivity(), TextView.OnEditorActionListener {
...
 
+        pQField.filters = arrayOf(NumInputFilter(this))
+        pQField.setOnEditorActionListener(this)
+
+        iFField = findViewById(R.id.image_format_setting_field)
+        iFField.setOnEditorActionListener(this)
+
+        vFField = findViewById(R.id.video_format_setting_field)
+        vFField.setOnEditorActionListener(this)
+
```
anonlatte, line 193: extract to another function like `initInputFields`
probably you may find another functionality which may be added there

anonlatte (changes requested):

app/src/main/java/app/grapheneos/camera/CamConfig.kt
```diff
@@ -55,10 +55,10 @@ private fun ExtensionsManager.isExtensionAvailableSafe(
     cameraSelector: CameraSelector,
     extensionMode: Int
 ): Boolean {
-    try {
-        return isExtensionAvailable(cameraSelector, extensionMode)
+    return try {
```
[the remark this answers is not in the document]
Zanthed: This is a temporary workaround for a CameraX bug that doesn't properly handle extensions so I think we'll instead remove it in the next CameraX release assuming they fix it.

PR #145, opened 2022-03-10
Add italian translation

unbranched:


unbranched:
`/home/runner/work/Camera/Camera/app/src/main/res/values-it/strings.xml:33:4: Failed to flatten XML for resource 'image_quality_desc' with error: Invalid unicode escape sequence in string`

I guess it wants `&agrave;` in place of "à"? Sorry for the mistake.

Zanthed:
We're interested in this but we need to make a translation team for the languages so that it can be reviewed, maintained and so we can ask for it to be updated after we change the strings. We'll ping you when we're ready to assemble a team for Chinese. We'll try to do this in the near future.

anonlatte:
@Zanthed you may setup something like [Weblate](https://weblate.org/ru/) which will help and automate translations. 
Cons: sometimes there will be a lot of PRs

thestinger:
The way we want to do it is probably figuring out the top languages used by our community and then making translation teams with each one having their own Matrix channel. Those rooms can be used to coordinate translating apps, and before we're going to do a release we can ping all of them and get it done. I don't think we want to introduce additional software. Once we're ready for that we can go back through these PRs and let people who tried to contribute know about it.

PR #154, opened 2022-03-26
keep screen on for few min while camera app is in foreground

empratyush:


thestinger:
Merged since it was also included in the other PR.

PR #186, opened 2022-04-19
yo how do i reach out to you! 

k0r1g:
Here's my twitter: https://twitter.com/korigrogers

Would love to chat with you, give me a DM! I co-founded MoonDAO and we have a market-cap of like $200M- reckon there are some fun things we could work on!


thestinger:
Will respond on Twitter.

PR #195, opened 2022-04-26
Add dutch translations

GersomR-afk:
🇳🇱🇳🇱🇳🇱


thestinger:
Needs to wait until after all strings are moved to resources and we set up an official translation team for the language. Please help with finishing up moving strings to resources. There's an open PR but it had some issues that haven't been addressed.

PR #211, opened 2022-05-10
Increase field of view for devices with default video stabilization

ASerbinski:
Digital video stabilization reduces the field of view to provide a
buffer to move the target area around within. This mode does not
make sense for static images, therefore deactivate this mode.

Reference: https://issuetracker.google.com/issues/230013960?pli=1

Signed-off-by: Adam Serbinski <adam@serbinski.com>

### DIFF
--- a/app/src/main/java/app/grapheneos/camera/CamConfig.kt
+++ b/app/src/main/java/app/grapheneos/camera/CamConfig.kt
@@ -7,6 +7,8 @@ import android.content.Context
 import android.content.SharedPreferences
 import android.graphics.Bitmap
 import android.graphics.ImageDecoder
+import android.hardware.camera2.CameraMetadata
+import android.hardware.camera2.CaptureRequest
 import android.media.MediaMetadataRetriever
 import android.net.Uri
 import android.os.Build
@@ -21,6 +23,7 @@ import android.view.animation.LinearInterpolator
 import android.widget.Button
 import android.widget.FrameLayout
 import androidx.annotation.StringRes
+import androidx.camera.camera2.interop.Camera2Interop
 import androidx.camera.core.AspectRatio
 import androidx.camera.core.Camera
 import androidx.camera.core.CameraSelector
@@ -1302,13 +1305,23 @@ class CamConfig(private val mActivity: MainActivity) {
             }
         }
 
-        preview = Preview.Builder()
+        var previewBuilder = Preview.Builder()
             .setTargetRotation(
                 preview?.targetRotation
                     ?: rotation
             )
-            .setTargetAspectRatio(aspectRatio)
-            .build()
+            .setTargetAspectRatio(aspectRatio);
+
+        @androidx.camera.camera2.interop.ExperimentalCamera2Interop
+        if (!isVideoMode) {
+            Camera2Interop.Extender(previewBuilder)
+                .setCaptureRequestOption(
+                    CaptureRequest.CONTROL_VIDEO_STABILIZATION_MODE,
+                    CameraMetadata.CONTROL_VIDEO_STABILIZATION_MODE_OFF
+                )
+        }
+
+        preview = previewBuilder.build()
 
         preview?.let {
             useCaseGroupBuilder.addUseCase(it)


Zanthed:
Drop the merge commit please

Zanthed:
There is also a lint issue with your code.

```
> Task :app:lintDebug FAILED
Lint found 4 errors, 14 warnings. First failure:

/home/runner/work/Camera/Camera/app/src/main/java/app/grapheneos/camera/CamConfig.kt:1316: Error: This declaration is opt-in and its usage should be marked with

The full lint text report is located at:
  /home/runner/work/Camera/Camera/app/build/intermediates/lint_intermediate_text_report/debug/lint-results-debug.txt

> Task :app:shrinkReleaseRes

127 actionable tasks: 127 executed
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:lintDebug'.
> Lint found errors in the project; aborting build.
```

Run `./gradlew lint` and see if you can solve the issue please

ASerbinski:
@Zanthed - I believe that should address your concerns.

ASerbinski:
Interested, perhaps. This PR was trivial and I was able to muddle my way through it despite my unfamiliarity with the project and with kotlin. I'll need some time to study the project before I can commit to it.

thestinger:
Could copy the Include audio setting and wire it up to this, which would already work, but then it needs code to query whether it's supported and to figure out the default value to set up the UI properly. The tricky part is figuring out how to query the current value especially since it could vary based on which camera is going to be used. Ideally CameraX would support this but it likely won't any time soon and we'll need to do it ourselves. Maybe CameraX has an API to get the Camera2 camera ID which is what would be ideal to query it via Camera2. I'm unfamiliar with the interop API and how to get the Camera2 provider handle from the CameraX provider, etc.

thestinger:
It's possible that it would make sense to support this feature for image capture too if it helps to provide stabilization for image capture which is likely. That means hard-wiring it to disabled for image capture isn't strictly an upgrade but rather is turning off a feature some people may find useful. We probably need to make this into a toggle and likely disable it by default for image capture.

ASerbinski:
I don't think that it would be meaningful to enable EIS for still images. In theory, it could make sense for additive multi-exposure images, such as night sight, and to a much lesser and likely negligible extent, HDR, however it would need to be implemented somewhere. What is clear, is that google themselves don't make use of it, since their own camera software uses the full field of view (therefore we can conclude that EIS is disabled for still images). Or, its possible that they are, and that the actual field of view is larger still than even what is presented by gcam. If google doesn't make use if it, then chances are they didn't expend the resources to implement it, and likely it wouldn't make sense to assume that such a benefit would be available unless we could determine that with some level of certainty.

thestinger:
It seems like it would still completely apply for still images and would help out a lot especially in low light situations where it's using longer exposure times. I don't know how it would interact with HDR+ and keep in mind HDR+ is being used on all Pixels since the Pixel 2 for our Camera app unless flash is being used.

thestinger:
It would be good to figure out in what situations HDR+ is used on the Pixel 6 based on logs perhaps and whether EIS helps with it, causes it to be disables, interferes with it, etc.

thestinger:
I switched `var` to `val` because the variable isn't reassigned (note it doesn't prevent modifying objects, just reassigning).

ASerbinski:
> It seems like it would still completely apply for still images and would help out a lot especially in low light situations where it's using longer exposure times. I don't know how it would interact with HDR+ and keep in mind HDR+ is being used on all Pixels since the Pixel 2 for our Camera app unless flash is being used.

Using the suggestion from the upstream bug report of ```dumpsys media.camera | grep -A1 "VideoStabilizationMode"``` from GCam, I'm not able to find ANY condition for still images in which it turns that mode to ON. It switches it on for video mode only. And yes, I even tried in extreme darkness.

I believe that there are two levels of cropping that is done. The still image mode that crops it just a little bit for additive multi-exposure shots (night sight, astrophotography, possibly HDR), and the huge crop mode for videos only.

thestinger:
We can probably release this as version 35 soon. There's just the possible issue of devices without support for this feature breaking if we try to set a value, even if the value is false. I don't think it will be an issue but you never know.

thestinger:
We could start by just adding a toggle for video stabilization to the video mode. It looks like the Camera2Interop API has what's needed to query the current value via camera info once we start video mode so we should be able to properly display the device's own default value without overriding it.

ASerbinski:
I just tested it on my Nexus 6, which has only OIS and not EIS. No crashing.

```
shell@shamu:/ $ dumpsys media.camera | grep "android\.control"                                                                                                     
      android.control.aeLockAvailable (10024): byte[1]
      android.control.awbLockAvailable (10025): byte[1]
      android.control.aeAvailableTargetFpsRanges (10014): int32[8]
      android.control.aeCompensationStep (10016): rational[1]
      android.control.availableVideoStabilizationModes (1001a): byte[1]
      android.control.maxRegions (1001c): int32[3]
      android.control.aeCompensationRange (10015): int32[2]
      android.control.availableEffects (10018): byte[9]
      android.control.availableSceneModes (10019): byte[15]
      android.control.sceneModeOverrides (1001d): byte[45]
      android.control.availableModes (10026): byte[3]
      android.control.aeAvailableAntibandingModes (10012): byte[4]
      android.control.afAvailableModes (10017): byte[5]
      android.control.awbAvailableModes (1001b): byte[9]
      android.control.aeAvailableModes (10013): byte[4]
      android.control.postRawSensitivityBoostRange (10027): int32[2]
      android.control.zoomRatioRange (1002e): float[2]
      android.control.aeLockAvailable (10024): byte[1]
      android.control.awbLockAvailable (10025): byte[1]
      android.control.aeAvailableTargetFpsRanges (10014): int32[4]
      android.control.aeCompensationStep (10016): rational[1]
      android.control.availableVideoStabilizationModes (1001a): byte[1]
      android.control.maxRegions (1001c): int32[3]
      android.control.aeCompensationRange (10015): int32[2]
      android.control.availableEffects (10018): byte[9]
      android.control.availableSceneModes (10019): byte[15]
      android.control.sceneModeOverrides (1001d): byte[45]
      android.control.availableModes (10026): byte[3]
      android.control.aeAvailableAntibandingModes (10012): byte[4]
      android.control.afAvailableModes (10017): byte[1]
      android.control.awbAvailableModes (1001b): byte[9]
      android.control.aeAvailableModes (10013): byte[2]
      android.control.zoomRatioRange (1002e): float[2]
```

PR #213, opened 2022-05-11
Add persistent toggle for EIS

ASerbinski:
When in Video mode, settings includes persistent toggle for EIS.
EIS, when enabled, will reduce the field of view.

### DIFF
--- a/app/src/main/java/app/grapheneos/camera/CamConfig.kt
+++ b/app/src/main/java/app/grapheneos/camera/CamConfig.kt
@@ -91,6 +91,7 @@ class CamConfig(private val mActivity: MainActivity) {
             const val VIDEO_QUALITY = "video_quality"
             const val ASPECT_RATIO = "aspect_ratio"
             const val INCLUDE_AUDIO = "include_audio"
+            const val INCLUDE_EIS = "include_eis"
             const val SCAN = "scan"
             const val SCAN_ALL_CODES = "scan_all_codes"
             const val SAVE_IMAGE_AS_PREVIEW = "save_image_as_preview"
@@ -128,6 +129,8 @@ class CamConfig(private val mActivity: MainActivity) {
             const val CAMERA_SOUNDS = true
 
             const val INCLUDE_AUDIO = true
+
+            const val INCLUDE_EIS = true
 
             const val SCAN_ALL_CODES = false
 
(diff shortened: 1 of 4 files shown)


ASerbinski:
Note that this does NOT implement any form of querying for the default state for the device, not does it query if the device even supports EIS to begin with, so I think it probably needs a little extra work before merge.

thestinger:
Yeah, we'll need that in order to merge it. I think it should be possible to do that with CameraX's Camera2Interop API using their camera info API. You can query whether it's supported and the default value if it's supported once the video mode is already started.

It's possible that it does make sense to expose this in image capture mode but it's a super niche feature and Google Camera doesn't expose super niche features since it's meant to be a point and shoot camera app rather than resembling a DSLR with a ton of configuration options. We eventually want to add DSLR-style configuration for exposure, focus, white balance, etc. but while preserving the simplicity of the interface for most users who wouldn't use those features.

ASerbinski:
Does the device default value actually matter? I doubt very much that GCam bothers with checking the default, rather just turns on video stabilization when in video mode unless the user has manually turned it off. That would create an application level default and improve consistency between devices running this software.

I think from my point of view, we just want to query whether video stabilization is supported at all, and use that to decide if the UI element will be visible.

ASerbinski:
Added a commit that checks if EIS is supported by the device to determine if the setting should be shown.

Tested on Pixel 6 Pro (EIS supported) and on Nexus 6 AOSP 11 (EIS not supported).

thestinger:
@ASerbinski The reason the default might matter is because some devices may implement EIS but with a very poor quality implementation. We could just enable it by default regardless.

thestinger:
@ASerbinski Can you rename the setting from "Include EIS" to "Enable EIS"? Include made sense for audio since it's about whether it's included in the capture or not.

thestinger:
Some devices may also have a broken implementation.

ASerbinski:
I'd say that if its a poor or broken implementation, it shouldn't be advertised as a feature to begin with.
I can implement that if you insist on it, but its probably overkill, and increases complexity with likely no substantial benefit.

I've made the change from include to enable, and did so throughout the code to keep it easy to follow.

thestinger:
I squashed it into 1 commit since the other history based on the review isn't going to be helpful to others.

Issue #204, opened 2022-05-02
QR Scan freezes camera


mikitakahazekura:
https://user-images.githubusercontent.com/24222032/166290230-dd8241c7-e8ed-4091-91fb-52604eaa38fa.mp4

Steps to reproduce:
1) Open Camera
2) Switch to QR Scan
3) Camera viewer freezes

Device: OnePlus 6
OS: DivestOS 18.1
Build Number: RQ3A.211001.001
Kernel: 4.9.282-dos.p132-gd57522d38ec1-dirty Sat Apr 9 14:12:58 EDT 2022
Logcat:
[logcat.txt](https://github.com/GrapheneOS/Camera/files/8604746/logcat.txt)

Zanthed:
This appears to be a library bug in Material and not our issue.

Zanthed:
@mikitakahazekura Please try with Camera v31.

thestinger:
Another user also reported an issue on the OnePlus 6. It seems that this is a OnePlus 6 OS bug that's likely also present on the stock OS.

thestinger:
@mikitakahazekura Please try again with version 36 or version 37 to see if the CameraX 1.1.0-rc01 update fixed this. It's likely still an issue but knowing for sure that's it's still an issue will help me report an issue to CameraX.

SkewedZeppelin:
I can still reproduce this issue on that device.

thestinger:
It's possible that the issue is caused by newer Android versions using Scudo instead of jemalloc for this process. It appears to be a memory corruption bug crashing the camera service which you can see in the log that was posted. It's definitely an OS issue but it's possible we could somehow work around it or CameraX could work around it. It would be extremely helpful to know if this issue occurs on the most recent stock OS release for the OnePlus 6 or 6T since if it doesn't then it's likely a bug which was in the stock OS but got uncovered by later AOSP changes.

If you look at the hardened_malloc repository, you can see the 12.1 branch has a workaround for Pixel 3 / Pixel 3 XL memory corruption bugs in the same generation of Qualcomm Camera service as this device. OnePlus 6 appears to be using a 32-bit process so hardened_malloc isn't relevant and it could be getting uncovered by Scudo instead, although Scudo doesn't uncover the issue on the Pixel 3 that's still present.

SkewedZeppelin:
I already apply the hmalloc camera exception to all camera libraries on DivestOS

thestinger:
[...]

PR #219, opened 2022-05-14
Add workaround for devices with broken region focus support

SkewedZeppelin:
In this case the OnePlus 6 and 6T

Closes https://github.com/GrapheneOS/Camera/issues/204

### DIFF
--- a/app/src/main/java/app/grapheneos/camera/ui/activities/MainActivity.kt
+++ b/app/src/main/java/app/grapheneos/camera/ui/activities/MainActivity.kt
@@ -191,10 +191,22 @@ open class MainActivity : AppCompatActivity(),
             previewView.width.toFloat(), previewView.height.toFloat()
         )
 
-        val autoFocusPoint = factory.createPoint(
-            previewView.width / 2.0f,
-            previewView.height / 2.0f, qrOverlay.size
-        )
+        val shouldUseFocusRegion = when (Build.PRODUCT) {
+            "OnePlus6", "OnePlus6T" -> false
+            else -> true
+        }
+
+        val autoFocusPoint = if (shouldUseFocusRegion) {
+            factory.createPoint(
+                previewView.width / 2.0f,
+                previewView.height / 2.0f, qrOverlay.size
+            )
+        } else {
+            factory.createPoint(
+                previewView.width / 2.0f,
+                previewView.height / 2.0f
+            )
+        }
 
         camConfig.camera?.cameraControl?.startFocusAndMetering(
             FocusMeteringAction.Builder(autoFocusPoint).disableAutoCancel().build()


app/src/main/java/app/grapheneos/camera/ui/activities/MainActivity.kt
```diff
@@ -203,6 +210,13 @@ open class MainActivity : AppCompatActivity(),
         startFocusTimer()
     }
 
+    private fun shouldUseFocusRegion(): Boolean {
+        //OnePlus 6/6T: https://github.com/GrapheneOS/Camera/issues/204
```
SkewedZeppelin, line 214: Is it worth it to cache this value?

app/src/main/java/app/grapheneos/camera/ui/activities/MainActivity.kt
```diff
@@ -191,10 +191,17 @@ open class MainActivity : AppCompatActivity(),
             previewView.width.toFloat(), previewView.height.toFloat()
         )
 
-        val autoFocusPoint = factory.createPoint(
-            previewView.width / 2.0f,
-            previewView.height / 2.0f, qrOverlay.size
-        )
+        val autoFocusPoint = if (shouldUseFocusRegion()) {
```
SkewedZeppelin, line 194: Is there a some cleaner Kotlin way to do this?

thestinger: I don't think there's a better way to do this part.

app/src/main/java/app/grapheneos/camera/ui/activities/MainActivity.kt
thestinger: You should be able to replace the function with a field. Can also use `when` for pattern matching since we're going to need to add the other OnePlus 6 and 6T model names in addition to these 2. I'm unsure where to get a full list. I could see if the Play Store has a list.

```
val shouldUseFocusRegion = when (Build.MODEL) {
    "ONEPLUS A6003", "ONEPLUS A6013" -> true
    else -> false
}
```

thestinger: Did that backwards, fixed.

thestinger: Is there a field we could use that's set to `OnePlus6` instead? Maybe Build.PRODUCT instead of Build.MODEL?

SkewedZeppelin: That works:
https://github.com/GrapheneOS/AttestationSamples/blob/main/ONEPLUS%20A6003/filtered.prop#L43
https://github.com/GrapheneOS/AttestationSamples/blob/main/ONEPLUS%20A6013/filtered.prop#L46

thestinger:
Going to merge this but we've confirmed it happens with the stock OS so we can get CameraX to add a quirk and then we can delete this code after we update CameraX to a new version. Hopefully we can get them to obtain one of the devices and add this for the June release, if not hopefully the July release. I'm unsure if they'd be willing to add the quirk without replicating it themselves.

SkewedZeppelin:
Thank you!

PR #223, opened 2022-05-17
improvements to app launch speed, gallery performance, secure activities, VideoCapturer

muhomorr:


MHShetty:
Hi @muhomorr,

Thanks a lot for making this PR! It includes a lot of helpful changes from handling unexpected scenarios (for e.g. showing a notification when an unexpected error occurs while saving an image) to improving various aspects of the app that in turn improves the overall UX to making the code much more efficient, readable and maintainable in the long run.

Here are a few issues/suggestions that I came across/up with while testing the latest changes:

The in-app gallery sometimes startups with a black background even after the image gets loaded while the app bar is still visible (not really sure if it is the immersive zooming support that we have added is causing this or maybe something unexpected that's happening with the latest changes). Unfortunately, I wasn't able to really come up with a definite way of reproducing this issue but trying to open the gallery several times eventually caused this issue to appear.

The primary purpose of storing the uris in shared preferences was to reduce the overall query time taken to open the gallery app. So bulk querying, each time the gallery is attempted to launch might be slower than directly fetching it from SharedPreferences. (We could try re-using the concept of EphemeralSharedPrefs (in-memory) to improve the overall speed of opening of the gallery)
https://github.com/GrapheneOS/Camera/pull/223/commits/c384facea59d277c9ebb5a354d6e36ff97efbb38

(Please feel free to ignore this) I tried testing the latest SAF changes with a storage provider, the new location that shows up is much more neater and legible compared to the existing implementation. Is there any way we could add the name of the storage provider to the path to make it easier to know which storage provider is being currently selected? (We could apply something to the name displayed in brackets)

I'll share about any more issue/suggestion(s) if I come across them while testing.

Thanks a lot for your valuable contribution @muhomorr!

Regards,
Mohit

muhomorr:
> The in-app gallery sometimes startups with a black background even after the image gets loaded while the app bar is still visible (not really sure if it is the immersive zooming support that we have added is causing this or maybe something unexpected that's happening with the latest changes). Unfortunately, I wasn't able to really come up with a definite way of reproducing this issue but trying to open the gallery several times eventually caused this issue to appear.

What do you mean by "app bar"? Black background should show up much rarer after new change that decodes first item and loads previously captured items concurrently.

> The SAF access seems to be non-persistent after the device restarts for some reason.

Unable to reproduce this.

> The primary purpose of storing the uris in shared preferences was to reduce the overall query time taken to open the gallery app. So bulk querying, each time the gallery is attempted to launch might be slower than directly fetching it from SharedPreferences. (We could try re-using the concept of EphemeralSharedPrefs (in-memory) to improve the overall speed of opening of the gallery) 

Previous code that tracked previously captured items was highly inefficient:
https://github.com/GrapheneOS/Camera/blob/ff6dc0e9a3d8159719a4c789c8daf7caeceb94d6/app/src/main/java/app/grapheneos/camera/CamConfig.kt#L516
Calling DocumentFile.exists() on each Uri is very slow when there's hundreds of them.
This led to multi-seconds delays in multiple places (mediaUris property is accessed when the MainActivity launches, when the gallery launches, and in few other places).
The reason it is slow is because each exists() call performs IPC, and permission checks are being performed during each of them. When Uri is a SAF Uri, permission checks involve one more IPC call for each exists() call.
Also:
- storing a string of unbounded size in shared preferences is a bad idea, it'll slow down initial parsing of SharedPreferences on app launch more and more as this string gets larger.
SAF Uris are long, each could easily be >100 bytes, which means 1 MB string after 10000 images.
- https://github.com/GrapheneOS/Camera/blob/ff6dc0e9a3d8159719a4c789c8daf7caeceb94d6/app/src/main/java/app/grapheneos/camera/CamConfig.kt#L544
`getOldPaths()` didn't work at all, because the app loses access to its previous files after reinstall (I assume this was the purpose of this function)
- https://github.com/GrapheneOS/Camera/blob/ff6dc0e9a3d8159719a4c789c8daf7caeceb94d6/app/src/main/java/app/grapheneos/camera/CamConfig.kt#L533
This code block is O(m * n), where m is size of the first list, and n is the size of the second. One of the lists should have been converted to a O(1) `Set` before performing this check. In practice, `DocumentFile.exists()` made this function very slow before this performance bug became an issue.
- I would strongly recommend against property getters that do any slow operations (especially IO). It's easy to forget that reading a property is not reading from a regular field, but performing slow IO calls.

Bulk querying of previously captured items has multiple advantages:
- very few IPC calls and permission checks
- MediaStore queries don't touch the file system, but come from its SQLite database, so they are fast, especially now that there's just one query
- most common SAF provider is the local filesystem provider (`com.android.externalstorage.documents`). When asking it to return the list of files in a directory, it's quite fast. All it does internally is call `File.listFiles()` and then perform few low-overhead operations on each item of this list before returning it
- slower SAF providers should work at least as fast with bulk querying as with previous approach

> I tried testing the latest SAF changes with a storage provider, the new location that shows up is much more neater and legible compared to the existing implementation. Is there any way we could add the name of the storage provider to the path to make it easier to know which storage provider is being currently selected?

I'll look into it. New changes improve this further by including the user-facing name of a storage volume, if it's available.

MHShetty:
Hi @muhomorr,

> What do you mean by "app bar"? Black background should show up much rarer after new change that decodes first item and loads previously captured items concurrently.

App bar as in the action bar (that consists of back, edit, delete, share and more actions) that shows up in the in-app gallery. I tried looking for a definite way to reproduce this issue but was unable to, it just randomly occurs while trying to open the in-app gallery multiple times.

Here's a screenshot of how the in-app gallery looks when this issue appears -

In-app gallery with black background

(Tapping the image or attempting to pan/zoom twice nrings the in-app gallery to the expected state)

> > The SAF access seems to be non-persistent after the device restarts for some reason.

> Unable to reproduce this.

I'm able to consistently reproduce on my Pixel 4a. Tried attempting it again on my device and its still the same. Not really sure what's causing this issue. Could be an issue with the SAF I used. Could you please share the name of the SAF you performed the testing with or any other custom SAF that's expected to work? (just in case if this is a device specific issue)

> getOldPaths() didn't work at all, because the app loses access to its previous files after reinstall (I assume this was the purpose of this function)

`getOldPaths()` was added to ensure that we don't miss out to show the media that was captured using an old method that we were previously using. This was done to ensure that the captured media doesn't disappear for the user instead of only relying on SharedPreferences (the current approach as you mentioned can get highly inefficient when the number of images increase over time).  Uninstalling and installing the app causes the images to be lost, but an update/reinstall (without uninstalling) still keeps the media captured by the app so it helped us back then in those cases.

Regarding the implementation for querying/retrieving the media captured by the app.. We had switched over to using SharedPreferences back then as MediaStore always had some fixed amount of delay time (irrespective of the number of media uris it loaded) and having two separate queries made it much more visible while trying to open the in-app gallery. SharedPreferences was used to make that faster/counter that, but as you said it could becoming a problem at a later stage when the user has too many photos/videos (the overhead for each media (checking if it exists etc.), and risk of having an unbounded list of elements etc.). Currently, the new UI flow designed in this PR makes the delay taken by MediaStore much less visible (thereby improving the overall UX of the in-app gallery).

I'll just keep occasionally testing the PR and let you know whenever I come across any more issues. I'll mostly be back with properly contributing (code, review, etc.) within the next 2 weeks.

Thanks a lot for the valuable time, feedback and contributions @muhomorr!
Hope to work with you more in the near future :)

Regards,
Mohit

thestinger:
I don't think we need to worry much about the legacy images from very early camera releases disappearing from the app's gallery at this point since they'll still be present in storage so it's not a big deal.

muhomorr:
> I don't think we need to worry much about the legacy images from very early camera releases disappearing from the app's gallery at this point since they'll still be present in storage so it's not a big deal.

They won't disappear. MediaStore keeps track of all images that were saved to the default location, and vast majority of users of SAF locations will be covered by the policy of keeping track of last 5 SAF "trees" (directories), including the ones that were picked by the previous versions of the app.

MHShetty:
Yes that's true. The approach that was introduced back then only relied on SharedPreferences and not MediaStore, hence the media captured before the update wouldn't be visible in the **in-app gallery**. The latest changes in this PR would resolve this as an issue by including all the images that were taken from the very beginning since MediaStore remembers them (assuming the user didn't uninstall the app at some point in time).

muhomorr:
> App bar as in the action bar (that consists of back, edit, delete, share and more actions) that shows up in the in-app gallery. I tried looking for a definite way to reproduce this issue but was unable to, it just randomly occurs while trying to open the in-app gallery multiple times.

> Here's a screenshot of how the in-app gallery looks when this issue appears -

Unable to reproduce this. Was this screenshot made on the build with the second batch of commits (pushed yesterday)?

Regarding persisted Uri permission grants:
I have tested with the `com.android.externalstorage.documents` provider on Pixel 4a, works well. Persisting of Uri permission is managed by the OS (see `com.android.server.uri.UriGrantsManagerService`). In Android 10, 128 persistable grants per package were allowed, 512 grants since Android 11. Tell me which storage provider doesn't work, I'll look into it.

> getOldPaths() was added to ensure that we don't miss out on showing the media that was captured using the old method that we were previously using. This was done to ensure that the captured media doesn't disappear for the user instead of only relying on SharedPreferences

I see. Still, querying them each time was a bad idea. They should have been copied to the SharedPreferences on first launch.

> Regarding the implementation for querying/retrieving the media captured by the app.. We had switched over to using SharedPreferences back then as MediaStore always had some fixed amount of delay time (irrespective of the number of media uris it loaded) and having two separate queries made it much more visible while trying to open the in-app gallery. 

Can you elaborate on the "fixed amount of delayed time" part?

MHShetty:
Hey @muhomorr,

I tried testing the latest v38 of our camera application and this issue no longer seems to persist there and the launching and finishing of the in-app gallery is much more smoother now.

> Unable to reproduce this. Was this screenshot made on the build with the second batch of commits (pushed yesterday)?

It was made on the first batch of commits back then. Possibly the reason for the issue persisting. Really sorry for the inconvinence caused here

> Regarding persisted Uri permission grant..

The latest release version (v38) seems to work as expected (and the debug version after pulling the latest changes) on my Pixel 4a. The issue might have got resolved in the second batch of commits that was made later (after the commit b61244da0c6938184a3c7b651932241e2e24f0f7). 

> I see. Still, querying them each time was a bad idea. They should have been copied to the SharedPreferences on first launch.

Yes that's true, copying it at once and possibly using another SharedPreferences key as a flag (to check whether the copying was successfully done before) would be an better approach here.

> Can you elaborate on the "fixed amount of delayed time" part?

There is a certain amount of delay that the in-app gallery takes to open irrespective of the number of images taken by the app (or that is being stored by MediaStore) [even if it's just a single image/video]. So was referring to that there while specifying the reason for shifting to the SharedPreferences approach we took. However, the other checks that we made (checking if each uri exists [the IPC call(s) it would make, as you earlier mentioned], etc.) could have made it slower in the long run.

> Thank you for your kind words and for the help with testing.

You're welcome!

MHShetty:
Hey @muhomorr,

I tried testing the latest v38 of our camera application and this issue no longer seems to persist there and the launching and finishing of the in-app gallery is much more smoother now.

> Unable to reproduce this. Was this screenshot made on the build with the second batch of commits (pushed yesterday)?

It was made on the first batch of commits back then. Possibly the reason for the issue persisting. Really sorry for the inconvinence caused here

> Regarding persisted Uri permission grant..

The latest release version (v38) seems to work as expected (and the debug version after pulling the latest changes) on my Pixel 4a. The issue might have got resolved in the second batch of commits that was made later (after the commit b61244da0c6938184a3c7b651932241e2e24f0f7). 

> I see. Still, querying them each time was a bad idea. They should have been copied to the SharedPreferences on first launch.

Yes that's true, copying it at once and possibly using another SharedPreferences key as a flag (to check whether the copying was successfully done before) would be an better approach here.

> Can you elaborate on the "fixed amount of delayed time" part?

There is a certain amount of delay that the in-app gallery takes to open irrespective of the number of images taken by the app (or that is being stored by MediaStore) [even if it's just a single image/video]. So I was referring to that there while specifying the reason for shifting to the SharedPreferences approach we took. However, the other checks that we made (checking if each uri exists [the IPC call(s) it would make, as you earlier mentioned], etc.) could have made it slower in the long run.

> Thank you for your kind words and for the help with testing.

You're welcome!

PR #240, opened 2022-05-26
German translation

Margu86:
added translation for values.xml and arrays.xml
made changes in SettingsDialog.kt to allow the usage of arrays.xml in different languages


thestinger:
This will need to be rebased to be a single commit. We're not ready to take this yet but it will be useful as a starting point in the future.

thestinger:
It will be helpful if you could submit changes for splitting out strings separately instead of as part of the translation. We're going to need to move all the strings into resources and make them translatable before we can start accepting translations. It's a bit of a mess right now. It would be easier to start with PDF Viewer.

Margu86:
For the PDF Viewer I already have an open translation PR ;-)
I will close this PR and create a new one, only including the translated strings.xml without any other changes.
So there is a clean starting point for the time all strings are moved.

@thestinger 
Hope this is okay with you?

thestinger:
Ideally submit changes moving strings to strings.xml, etc. first since it will be needed for all translations and is easy for us to review and merge before we put together translation teams able to review stuff in languages most of our developers don't know.

Margu86:
Alright string moving it is then

PR #249, opened 2022-06-04
Safely allow multiple image capture requests to be serially processed

MHShetty:


thestinger:
Don't want to take this approach.

PR #251, opened 2022-06-05
Move strings to resources

Margu86:


app/src/main/java/app/grapheneos/camera/CamConfig.kt
```diff
@@ -721,7 +722,7 @@ class CamConfig(private val mActivity: MainActivity) {
         }
 
         if (!commonPref.contains(SettingValues.Key.FOCUS_TIMEOUT)) {
-            editor.putString(SettingValues.Key.FOCUS_TIMEOUT, SettingValues.Default.FOCUS_TIMEOUT)
+            editor.putString(SettingValues.Key.FOCUS_TIMEOUT, mActivity.resources.getStringArray(R.array.time_options)[SettingValues.Default.FOCUS_TIMEOUT])
```
thestinger, line 725: This is going to be broken. The storage needs to be fixed to use separate values from the user-facing strings. It was the wrong way to store these settings but not much can be done about that beyond splitting it up.

thestinger:
Can't be merged as is.

PR #254, opened 2022-06-08
Add support for continuous shots

MHShetty:


app/src/main/java/app/grapheneos/camera/capturer/ImageCapturer.kt
```diff
@@ -29,26 +29,32 @@ var isTakingPicture: Boolean = false
...
         mActivity.captureButton.isEnabled = true
 
-        val animation: Animation = AlphaAnimation(mActivity.captureButton.alpha, 1f)
-        animation.duration = 200
-        animation.interpolator = LinearInterpolator()
-        animation.fillAfter = true
+        cFadeAnim?.cancel()
+
+        cFadeAnim = AlphaAnimation(mActivity.captureButton.alpha, 1f)
+        cFadeAnim?.duration = 200
```
muhomorr, line 53: There's no need for null-safe `?.` here, it's guaranteed that `AlphaAnimation` is non-null.
Put it in a local `val`, or better yet, rewrite it like this:

```
        val button = mActivity.captureButton
        button.isEnabled = true

        cFadeAnim?.cancel()

        AlphaAnimation(button.alpha, 1f).apply {
            duration = 200
            interpolator = LinearInterpolator()
            fillAfter = true

            button.startAnimation(this)
            cFadeAnim = this
        }
```

Same applies to `fadeCaptureButton()`.

app/src/main/java/app/grapheneos/camera/ui/activities/MainActivity.kt
```diff
@@ -724,6 +724,18 @@ open class MainActivity : AppCompatActivity(),
         }
 
         captureButton = binding.captureButton
+        captureButton.setOnLongClickListener {
+            if (camConfig.isVideoMode || camConfig.isQRMode)
+                return@setOnLongClickListener false
+
+            if (timerDuration != 0) {
+                showMessage("Unable take consecutive shots when timer is set")
```
muhomorr, line 732: I don't think there's a need to show this message.

app/src/main/java/app/grapheneos/camera/capturer/ImageCapturer.kt
muhomorr: > The cFadeAnim is expected to be null when the first fade/unfade animation gets performed. We perform the cancel operation to avoid having multiple animations run on the same view while taking consecutive shots.

I understand that.
I meant that `cFadeAnim` is guaranteed not to be null at that line (it was just assigned) . Kotlin compiler doesn't know that `cFadeAnim` is not being modified by other threads, and thus requires null checks on each access.

An alternative would be the following:
```
val anim = AlphaAnimation(button.alpha, 1f)
anim.duration = 200
...
cFadeAnim = anim
```

MHShetty: > This comment was attached to line 53 (cFadeAnim?.duration = 200), not to line 50 (cFadeAnim?.cancel()).

Oh that's right, multiple null-checks could have been avoided there.

Have just pushed the required changes with the help of the `apply` method/approach.

app/src/main/java/app/grapheneos/camera/capturer/ImageCapturer.kt
```diff
@@ -36,25 +36,29 @@ class ImageCapturer(val mActivity: MainActivity) {
 
         cFadeAnim?.cancel()
 
-        cFadeAnim = AlphaAnimation(mActivity.captureButton.alpha, 0.6f)
-        cFadeAnim?.duration = 200
-        cFadeAnim?.interpolator = LinearInterpolator()
-        cFadeAnim?.fillAfter = true
+        AlphaAnimation(mActivity.captureButton.alpha, 0.6f).apply {
+            this.duration = 200
```
muhomorr, line 40: `this` can be omitted on lines 40..42 and 55..57.
If you want to keep it for clarity (and for avoiding potential naming conflicts, which I support), replace `apply` with `let` and `this` with `it`.

MHShetty: Yes using `let` should be a better option here since `this` could be confused with the instance of the main class (although had thought of skipping this then since `apply` is generally expected to be used in such cases)

app/src/main/res/values/strings.xml
```diff
@@ -185,4 +185,5 @@
 
     <string name="zsl_setting_title">Use ZSL in Latency mode</string>
     <string name="zsl_setting_desc">Uses Zero Shutter Lag (ZSL) in Latency mode for faster capture. Certain devices may have a buggy implementation for this.</string>
+    <string name="image_queue_full">The image queue is currently full.. Please wait for a while before taking your next set of shots.</string>
```
quh4gko8, line 188: Double dot typo here

app/src/main/java/app/grapheneos/camera/capturer/ImageSaver.kt
```diff
@@ -84,6 +89,7 @@ class ImageSaver(
         }
 
         imageWriterExecutor.execute(this::saveImage)
+        Log.i("TAG", "Trying to save image $photosInQueue")
```
muhomorr, line 92: Don't leave verbose logging in release builds.
Use `if (BuildConfig.DEBUG)` or similar check.
Also, reference `TAG` constant, not `"TAG"` string.
Same applies to line 132.

app/src/main/java/app/grapheneos/camera/capturer/ImageSaver.kt
```diff
@@ -123,7 +129,8 @@ class ImageSaver(
             handleError(e)
             return
         }
-
+        Log.i("TAG", "Done with image $photosInQueue")
+        --photosInQueue
```
muhomorr, line 133: It's incorrect to modify a regular variable from multiple threads without synchronization.
Line 78 is executed on the main thread, this line -- on a background thread.

app/src/main/java/app/grapheneos/camera/capturer/ImageSaver.kt
```diff
@@ -337,7 +345,10 @@ class ImageSaver(
         val imageCaptureCallbackExecutor = Executors.newSingleThreadExecutor()
         private val imageWriterExecutor = Executors.newSingleThreadExecutor()
 
+        private var photosInQueue = 0
+
         private const val TAG = "ImageSaver"
+        private const val MAX_PHOTOS_IN_QUEUE = 20
```
muhomorr, line 351: 20 is way too high of a limit. Depending on resolution, it can require from around 100 MiB (5 MiB per ~12 Mpx image) to close to 200 MiB of memory.
I think 5 is enough.

app/src/main/java/app/grapheneos/camera/capturer/ImageSaver.kt
```diff
@@ -337,7 +345,10 @@ class ImageSaver(
         val imageCaptureCallbackExecutor = Executors.newSingleThreadExecutor()
         private val imageWriterExecutor = Executors.newSingleThreadExecutor()
 
+        private var photosInQueue = 0
```
muhomorr, line 348: This variable needs to be of `AtomicInteger` type for thread safety.

Also, it's updated in wrong places.
Line 326 is called only after image capture errors, not image saving errors.
Line 133 may will be reached if image saving fails.
` ++photosInQueue` on line 78 is misleading, no image is added to an `ImageSaver` queue at that point, en queuing happens only after successful `onCaptureSuccess` callback.


A correct implementation would be as follows:
`private val imageWriterQueueLength = AtomicInteger()`

`imageWriterQueueLength.getAndIncrement()` right before `imageWriterExecutor.execute(this::saveImage)`
and
`imageWriterQueueLength.getAndDecrement()` at the very beginning of `saveImage()` function.

app/src/main/java/app/grapheneos/camera/capturer/ImageCapturer.kt
```diff
@@ -32,22 +31,27 @@ class ImageCapturer(val mActivity: MainActivity) {
...
             cFadeAnim = it
         }
     }
 
     private fun unfadeCaptureButton() {
-        mActivity.captureButton.isEnabled = true
+
+        val captureButton = mActivity.currentCaptureButton
+
+        captureButton.isEnabled = true
```
muhomorr, line 54: Lines 58 and 63 need to be updated too.

muhomorr (changes requested):

app/src/main/java/app/grapheneos/camera/ui/SettingsDialog.kt
```diff
@@ -77,6 +77,10 @@ class SettingsDialog(val mActivity: MainActivity) :
     private var videoQualitySetting: View
     private var timerSetting: View
 
+    var smRadioGroup: RadioGroup
```
muhomorr, line 80: These names are too short. Also, `var` should be replaced with `val` whenever possible.

app/src/main/res/values/strings.xml
MHShetty: Hey @quh4gko8,

Actually that was kind of intended, although this message will most likely be replaced with something else after a few more changes are pushed to the branch. Thanks for the feedback though @quh4gko8!

app/src/main/java/app/grapheneos/camera/capturer/ImageSaver.kt
MHShetty: Hey @muhomorr,

Actually those log checks were only present to make debugging the app a bit easier while testing/reviewing. I'll just remove these logs soon. Thanks for mentioning about them @muhomorr!

app/src/main/java/app/grapheneos/camera/ui/SettingsDialog.kt
MHShetty: I'll make these changes (for the `SettingsDialog` class as a whole) to the main branch of our main repository separately once we're done merging these changes. Thanks a lot for making a note of this @muhomorr!

app/src/main/java/app/grapheneos/camera/capturer/ImageSaver.kt
MHShetty: I'll soon replace the regular `Int` variable with `AtomicInteger` and its other usages to ensure thread safety.

app/src/main/java/app/grapheneos/camera/capturer/ImageCapturer.kt
MHShetty: I think there's probably some GitHub UI issue here. Could you please clarify on this review once?
Thanks a lot for your valuable time and feedback @muhomorr!

app/src/main/java/app/grapheneos/camera/capturer/ImageSaver.kt
muhomorr: What do you mean?
In vast majority of cases, actual queue length will never exceed 2, because image writeback usually is much faster than image capture.
This check is needed only for rare case of slow storage / slow SAF storage provider.
Setting this value to 10 will mean that the user can lose up to 10 last captured images in case app crashes / storage provider becomes unavailable / storage becomes completely full / etc

app/src/main/java/app/grapheneos/camera/capturer/ImageCapturer.kt
muhomorr: I mean lines 58 and 63 still refer to `mActivity.captureButton` (not the same as `captureButton`)

muhomorr: See changes on lines 40 and 45

app/src/main/java/app/grapheneos/camera/capturer/ImageSaver.kt
MHShetty: Yes, we had introduced this change mainly for slow SAFs and other disk storages.

> Setting this value to 10 will mean that the user can lose up to 10 last captured images in case app crashes / storage provider becomes unavailable / storage becomes completely full / etc.

Losing the captured photos could be a good reason to reduce the limit to 5 over being able to take multiple shots (over 10) in case of slow storage providers.

Will make the required changes soon. Thanks for the feedback @muhomorr!

app/src/main/java/app/grapheneos/camera/capturer/ImageSaver.kt
muhomorr: > photosInQueue variable represents the image capture requests to be both captured

Why keep track of image capture requests? There's never more than one. Name of variable `photosInQueue` suggests that it relates to the length of the queue. Image capture requests are not queued by the app, they are dispatched right away.

> The issue with just keeping it the way you are suggesting/the approach I had taken before actually pushing the changes is that there's a small possibility that the next ImageSaver instance can be instantiated before the increment operation takes place causing an extra image capture request to be added when the queue is already full based on the max limit we have set

I'd just rename `MAX_PHOTOS_IN_QUEUE` to `MAX_LENGTH_OF_IMAGE_WRITER_QUEUE`. Keeping code as simple as possible is a good idea, especially for code that handles rare edge cases. In vast majority of cases (I'd say 9999 out of 10000), this limit will not be reached.

> We expect the queue length to decrement after the save operation completes (irrespective of the outcome [i.e. success/failure]) for the same reason.

In any case, current code doesn't handle image saving errors. `onError` callback isn't invoked in that case.

MHShetty: > Why keep track of image capture requests? There's never more than one.

The image capture request there involves both the actual capture and saving (as mentioned).

There is a possibility that the queue limit of such requests exceeds 

To reproduce this condition (i.e. in case of slow SAFs), we could add some delay below the line that records the timestamp of start of writing...
https://github.com/GrapheneOS/Camera/blob/d8fdcb9bac05930af8c6e88ed0eb5763c728d569/app/src/main/java/app/grapheneos/camera/capturer/ImageSaver.kt#L155-L156
...on line 156 (possibly using `TimeUnit.SECONDS.sleep(2L)`) and then try taking continuous shots until the queue gets full and once you see the snackbar message try making another set of continuous, while observing the length of the queue (using the log statements that were discussing about earlier in this thread). 

After making the `AtomicInteger` related changes, I'm still able to reproduce this possibility (starts from 1 and goes right upto 21 [as per the current limits]). To understand the issue better we could replace the `>=` check with `==`. The request queue would exceed indefinitely if image capture requests (from the user) are further made assuming we only stick to tracking the capture requests and not tracking both. Please try testing this out once.

> I'd just rename MAX_PHOTOS_IN_QUEUE to MAX_LENGTH_OF_IMAGE_WRITER_QUEUE. Keeping code as simple as possible is a good idea, especially for code that handles rare edge cases. In vast majority of cases (I'd say 9999 out of 10000), this limit will not be reached.

Image writer queue might be a bit misleading based on the current intended usage.

> In any case, current code doesn't handle image saving errors. onError callback isn't invoked in that case.

The decrement operation is performed irrespective of the outcome of the image saver. The decrement in the `onError` callback is to handle cases when the capture fails.

```
    private fun saveImage() {
        // Save operation starts
        try {
            saveImageInner()
        } catch (e: ImageSaverException) {
            handleError(e)
            return
        }
        // Save operation completes
        Log.i("TAG", "Done with image $photosInQueue")

        // The decrement operation gets performed indicating that one request has been completely processed
        photosInQueue.getAndDecrement()

        imageCapturer.mActivity.thumbnailLoaderExecutor.executeIfAlive(this::generateThumbnail)
    }
```

app/src/main/java/app/grapheneos/camera/capturer/ImageCapturer.kt
MHShetty: I'm really sorry but am unable to get you here. We're loading the current capture button to a local `captureButton` variable so we're ideally always referring to the instance returned by the `currentCaptureButton` getter field for a given method invocation.

app/src/main/java/app/grapheneos/camera/capturer/ImageSaver.kt
muhomorr: I don't think it matters if queue length briefly exceeds MAX_PHOTOS_IN_QUEUE constant by 1.

It's true that `if (photosInQueue.get() == MAX_PHOTOS_IN_QUEUE)` needs to be replaced with `if (photosInQueue.get() >= MAX_PHOTOS_IN_QUEUE)` to handle this case, I missed that at first.

> The decrement operation is performed irrespective of the outcome of the image saving operation

This is incorrect.
Follow the control flow:
`saveImage()` -> `saveImageInner()` -> `saveImageInner()` throws an `ImageSaverException` ->
`handleError()` and `return` are executed.

Lines
```
 // Save operation completes
        Log.i("TAG", "Done with image $photosInQueue")
   // The decrement operation gets performed indicating that one request has been completely processed
        photosInQueue.getAndDecrement()
```
are never reached in that case.

See it for yourself:
add
```
        if (true) {
            throw ImageSaverException(Place.FILE_CREATION, IllegalStateException())
        }
```
 at the beginning of `saveImageInner`
and observe that after 5 attempts at image capture, "queue is full" message will be shown each time, because `photosInQueue` will never be decremented, only incremented.

This is why I keep insisting on simplifying queue limiting code as much as possible, it's easy to get wrong.

app/src/main/java/app/grapheneos/camera/capturer/ImageCapturer.kt
muhomorr: > We're setting the current capture button to local variable captureButton, so we're ideally always referring to the instance returned by the currentCaptureButton getter field for a given method invocation.

Yes, this is correct. And `fadeCaptureButton()` was correctly updated to use `captureButton` local variable.
But `unfadeCaptureButton()` wasn't fully updated.
Line 54 correctly refers to `captureButton` local variable.
But lines
58 `AlphaAnimation(mActivity.captureButton.alpha, 1f).let {`
and 
63 `mActivity.captureButton.startAnimation(it)`

refer to `mActivity.captureButton` variable which is different from `captureButton` local variable.

app/src/main/java/app/grapheneos/camera/capturer/ImageSaver.kt
MHShetty: > I don't think it matters if queue length briefly exceeds MAX_PHOTOS_IN_QUEUE constant by 1.

Although it doesn't matter in terms of the reason, it still makes the code being pushed misleading, ideally the queue should respect its bounds.

> It's true that if (photosInQueue.get() == MAX_PHOTOS_IN_QUEUE) needs to be replaced with if (photosInQueue.get() >= MAX_PHOTOS_IN_QUEUE) to handle this case, I missed that at first.

Same here, this is still an workaround that can and has be avoided.

> This is incorrect.
> Follow the control flow: saveImage() -> saveImageInner() -> saveImageInner() throws an ImageSaverException ->
handleError() and return are executed.

The `return` statement here (causes the control to leave the method). Didn't notice this earlier.. thanks for the clarification.

> This is why I keep insisting on simplifying queue limiting code as much as possible, it's easy to get wrong.

It's already simple, it's just that the approach that the code that has been pushed takes is to consider the entire processing before removing an image from the queue (to ensure that the queue doesn't exceed it's limit). It's just that our approaches are different. (https://github.com/GrapheneOS/Camera/pull/254/commits/6e22196731c88e7142b3735ece111a33883a6052)

Thanks for raising out that concern @muhomorr!

app/src/main/java/app/grapheneos/camera/capturer/ImageCapturer.kt
MHShetty: Got it. Thanks for the feedback @muhomorr!

app/src/main/java/app/grapheneos/camera/capturer/ImageSaver.kt
```diff
@@ -126,13 +126,12 @@ class ImageSaver(
     private fun saveImage() {
         try {
             saveImageInner()
+            imageCapturer.mActivity.thumbnailLoaderExecutor.executeIfAlive(this::generateThumbnail)
```
muhomorr, line 128: This is bad code style.
Code that doesn't throw exceptions shouldn't be moved into `try {}` block.
Add a `finally{}' block and move `photosInQueue.getAndDecrement()` there.

I still disagree with your approach of tracking the queue length, but don't wan't to argue about it anymore.

MHShetty: > I still disagree with your approach of tracking the queue length, but don't want to argue about it anymore.

The approach you're asking me to take causes the queue to have more images than it otherwise should. I have already explained why that approach is being used already a couple of time earlier. We can ignore it and have some workarounds, but that still leaves the code broken.

MHShetty: > I still disagree with your approach of tracking the queue length, but don't want to argue about it anymore.

The approach you're asking me to take causes the queue to have more images than it otherwise should. I have already explained why that approach is being used already a couple of time earlier. We can ignore it and have some workarounds, but that still leaves the code broken.

The finally block has been added to the code. Thanks for your valuable time @muhomorr.

app/src/main/java/app/grapheneos/camera/capturer/ImageSaver.kt
```diff
@@ -126,13 +126,14 @@ class ImageSaver(
...
         } catch (e: ImageSaverException) {
             handleError(e)
-            return
+        } finally {
+            photosInQueue.getAndDecrement()
         }
         Log.i("TAG", "Done with image $photosInQueue")
-        photosInQueue.getAndDecrement()
-        imageCapturer.mActivity.thumbnailLoaderExecutor.executeIfAlive(this::generateThumbnail)
```
muhomorr, line 127: This is not what I meant. 
`imageCapturer.mActivity.thumbnailLoaderExecutor.executeIfAlive(this::generateThumbnail)` should stay outside the `try {}` block, because it doesn't throw `ImageSaverException` (or any exception, really)

MHShetty: Oh got it. Just pushed the changes. Thanks for the clarification!

PR #268, opened 2022-07-22
Build against android 13 APIs and migrate most deprecated APIs

quh4gko8:
`Linkify.ALL` alternative, `TextClassifier.generateLinks` is currently being looked if safe to call, replace the current implementation at https://github.com/repository-staging/Camera/blob/61506b061089e1ca1a19c4ed129054a9d81d59bf/app/src/main/java/app/grapheneos/camera/ui/activities/MainActivity.kt#L1103-L1104


thestinger:
Applying manually.

PR #293, opened 2022-10-04
[bugfixes] don't create intent chooser manually

### LINKED ISSUE #289: Missing option for storing default foto-editor
Pixel 6a, latest OS build (as of 2022-09-23)

GIVEN: shot an image with GrapheneOS camera app which is stored under the dedault folder /DCIM/Camera

WHEN I open the image with GrapheneOS camera app
AND I push in the edit shortcut 
AND the editor app shows all associated apps for editing this image

THEN there is no option to store the default foto-editor app (has to be choosen manually each time)

empratyush:
System will show a list of installed app that can handle this action and file type and user can select default app for this action which will work for all other installed app too.

fixes #289

### DIFF
--- a/app/src/main/java/app/grapheneos/camera/ui/activities/InAppGallery.kt
+++ b/app/src/main/java/app/grapheneos/camera/ui/activities/InAppGallery.kt
@@ -5,6 +5,7 @@ import android.animation.ValueAnimator
 import android.annotation.SuppressLint
 import android.app.Activity
 import android.app.AlertDialog
+import android.content.ActivityNotFoundException
 import android.content.Context
 import android.content.Intent
 import android.graphics.Color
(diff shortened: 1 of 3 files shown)


app/src/main/java/app/grapheneos/camera/ui/activities/InAppGallery.kt
```diff
@@ -204,9 +208,15 @@ class InAppGallery : AppCompatActivity() {
         editIntent.data = curItem.uri
 
         editIntent.flags = Intent.FLAG_GRANT_READ_URI_PERMISSION
-        editIntentLauncher.launch(
-            Intent.createChooser(editIntent, getString(R.string.edit_image))
-        )
+        if (withDefault){
```
muhomorr, line 211: Missing space before `{`

PR #320, opened 2022-11-03
do not proceed with drawing grids if mActivity is not initialised

octocorvus:
This fixes 'lateinit property mActivity has not been initialized' render error in android studio layout editor; this causes views to not render in the layout editor, thus making editing layout harder. On a real device mActivity should always be initialised.


octocorvus:
not required anymore, since layout editor in Android Studio Electric Eel and later versions works fine without this

Issue #82, opened 2022-01-02
Using ContraintLayout Instead of Nested Linear Layouts


4shutosh:
Nested Views are slow and poorly performant.

We should always try to keep layouts single hierarchical. This is best possible using Constraint Layout.

`activity_main`, `more_settings` are some of the examples where we can work on this.

PR #321, opened 2022-11-03
[UNTESTED] layout improvements

octocorvus:
This is part of effort to migrate `main_activity.xml` to contraint layout #82. It includes the following changes:

- remove `main_frame` and replace it with `preview_container`
- do not explicitly set `c_timer` (countdown timer) height and let constraints take care of that
- center gyroscope suggestions to the preview
- use preview grid's height instead of preview's height to calculate the area on which to draw the preview grid, since preview
 grid's height is the same as preview's height
- use `match_contraint` instead of `match_parent` in `preview_container` as `match_parent` is not recommended practice for constraint layouts


octocorvus:
Needs to be redone

PR #337, opened 2023-01-10
hide content in Recent Apps

Husseinfo:
Hide Camera content (last frame) when browsing Recent Apps using [FLAG_SECURE](https://developer.android.com/reference/android/view/WindowManager.LayoutParams#FLAG_SECURE).

Diff


thestinger:
This has undesirable consequences.

Husseinfo:
> This has undesirable consequences.

Can you elaborate please? Is it a technical thing?

thestinger:
This will disable screenshots which we don't want.

PR #355, opened 2023-04-03
Fix bytesToHex()

### LINKED ISSUE #354: Binary tab in QR Scan Mode is broken
https://github.com/GrapheneOS/Camera/blob/71462d8762f7f8c0217be4f09fe22bab6c7bf8a3/app/src/main/java/app/grapheneos/camera/ui/activities/MainActivity.kt#L1060-L1071

The output string uses at least 3 characters per byte of input, so the output length shouldn't be `bytes.size`.
Furthermore, the loop shouldn't be skipping over input bytes with `step 3` nor excluding the last byte.

I might submit a PR to fix this if I get the time.

delitako:
Fixes #354

### DIFF
--- a/app/src/main/java/app/grapheneos/camera/ui/activities/MainActivity.kt
+++ b/app/src/main/java/app/grapheneos/camera/ui/activities/MainActivity.kt
@@ -1058,16 +1058,37 @@ open class MainActivity : AppCompatActivity(),
     private lateinit var dialog: Dialog
 
     open fun bytesToHex(bytes: ByteArray): String {
-        val hexChars = CharArray(bytes.size)
-        for (i in 0 until bytes.indices.last - 1 step 3) {
-            val v: Int = bytes[i].toInt() and 0xFF
-            hexChars[i] = hexArray[v ushr 4]
-            hexChars[i + 1] = hexArray[v and 0x0F]
-            hexChars[i + 2] = ' '
+        if (bytes.isEmpty()) return "" // outLen will be wrong for empty inputs
+
+        // Represent bytes as a grid of hex digits:
+        // Add a space between every byte
+        // Double space every 4 bytes (unless end or newline)
+        // Add a newline every 8 bytes (unless end)
+
+        var outLen = bytes.size * 3 - 1 // 2 hex digits + 1 space/newline per byte (except last)
+        outLen += bytes.size / 8 // One double space per row except the last incomplete row
+        if (bytes.size % 8 > 4) {
+            outLen += 1 // One double space for the last incomplete row, if it has >4 columns
         }
+
+        val hexChars = CharArray(outLen)
+        var j = 0 // Output index
+
+        for (i in bytes.indices) {
+            val byte = bytes[i].toInt() and 0xFF
+            hexChars[j++] = hexArray[byte ushr 4]
+            hexChars[j++] = hexArray[byte and 0x0F]
+
+            if (i == bytes.lastIndex) break // No trailing whitespace
+            if (i % 8 == 7) {
+                hexChars[j++] = '\n'
+            } else {
+                hexChars[j++] = ' '
+                if (i % 4 == 3) hexChars[j++] = ' '
+            }
+        }
+
         return String(hexChars)
-            .replace("(.{12})".toRegex(), "$0  ")
-            .replace("(.{28})".toRegex(), "$0\n")
     }
 
     private var isQRDialogShowing = false


app/src/main/java/app/grapheneos/camera/ui/activities/MainActivity.kt
```diff
@@ -1058,16 +1058,35 @@ open class MainActivity : AppCompatActivity(),
...
+        // Add a newline every 8 bytes (unless end)
+
+        var outLen = bytes.size * 3 // 2 hex digits + 1 space/newline per byte
+        outLen += bytes.size / 4 // Double spaces
+        outLen -= bytes.size / 8 // No double space when newline
+        outLen -= 1 // No trailing space/newline
+        if (bytes.size % 4 == 0 && bytes.size % 8 != 0)
+            outLen -= 1 // No trailing double space
+
+        val hexChars = CharArray(outLen)
```
octocorvus, line 1072: for an empty `ByteArray`, `outLen` will be `-1`, which will cause exception when `CharArray` is initialised (might crash the app, if not handled). early returning an empty string (if the size of `ByteArray` is zero) can mitigate this

app/src/main/java/app/grapheneos/camera/ui/activities/MainActivity.kt
```diff
@@ -1058,16 +1058,35 @@ open class MainActivity : AppCompatActivity(),
...
+        // Add a space between every byte
+        // Double space every 4 bytes (unless end or newline)
+        // Add a newline every 8 bytes (unless end)
+
+        var outLen = bytes.size * 3 // 2 hex digits + 1 space/newline per byte
+        outLen += bytes.size / 4 // Double spaces
+        outLen -= bytes.size / 8 // No double space when newline
+        outLen -= 1 // No trailing space/newline
+        if (bytes.size % 4 == 0 && bytes.size % 8 != 0)
+            outLen -= 1 // No trailing double space
```
octocorvus, line 1070: this can be simplified to:
```kt
var outLen = bytes.size * 3 - 1 // 2 hex digits + 1 space/newline per byte (except last)
outLen += bytes.size / 8 // One double space per row except the last incomplete row
if (bytes.size % 8 > 4) {
    outLen += 1 // One double space for the last incomplete row, if it has >4 columns
}
```
also, note the usage of braces for `if` statement, read [android style guide](https://developer.android.com/kotlin/style-guide#braces) for more info

app/src/main/java/app/grapheneos/camera/ui/activities/MainActivity.kt
```diff
@@ -1058,16 +1058,35 @@ open class MainActivity : AppCompatActivity(),
...
     open fun bytesToHex(bytes: ByteArray): String {
-        val hexChars = CharArray(bytes.size)
-        for (i in 0 until bytes.indices.last - 1 step 3) {
-            val v: Int = bytes[i].toInt() and 0xFF
-            hexChars[i] = hexArray[v ushr 4]
-            hexChars[i + 1] = hexArray[v and 0x0F]
-            hexChars[i + 2] = ' '
+        // Add a space between every byte
+        // Double space every 4 bytes (unless end or newline)
+        // Add a newline every 8 bytes (unless end)
```
octocorvus, line 1063: (optional): it might be better to make it clear that the output is formatted as a matrix

octocorvus (changes requested):

app/src/main/java/app/grapheneos/camera/ui/activities/MainActivity.kt
```diff
@@ -1058,16 +1058,38 @@ open class MainActivity : AppCompatActivity(),
...
+
+        var outLen = bytes.size * 3 - 1 // 2 hex digits + 1 space/newline per byte (except last)
+        outLen += bytes.size / 8 // One double space per row except the last incomplete row
+        if (bytes.size % 8 > 4) {
+            outLen += 1 // One double space for the last incomplete row, if it has >4 columns
         }
+
+        // outLen will be wrong for empty inputs
+        if (bytes.isEmpty()) return ""
```
octocorvus, line 1073: Can you move this higher? Ideally, it should return before `outLen` is calculated, since it doesn't make much sense to calculate `outLen` if `bytes` is empty.

octocorvus (changes requested):

app/src/main/java/app/grapheneos/camera/ui/activities/MainActivity.kt
delitako: Yeah I thought the code flowed better this way even though calculating `outLen` first is useless. I'll change it though.

octocorvus (approved):
looks good to me

Issue #51, opened 2021-11-05
Add setting option for video rate


emersion:
Some cameras (e.g. Pixel 3a) support video capture up to 240Hz (with a lower resolution). This is useful in some niche use-cases, e.g. timing events (physics, electronics, etc).

thestinger:
I think this depends on CameraX adding support for it. Ultra high frame rate modes may require a CameraX extension instead of simply changing the rate. Not sure about that.

MHShetty:
I have made a feature request for this, it's status can be tracked from here: 
https://issuetracker.google.com/issues/205334816

Donkey-Doug:
What about resolution? My phone (Xiaomi Mi 9T Pro) only films in 1080p @ 30fps.

MHShetty:
Hi @Donkey-Doug,

> What about resolution? My phone (Xiaomi Mi 9T Pro) only films in 1080p @ 30fps.

We might need to wait until the CameraX team includes this in their public schedule or any of their future release to be able to actually alter these configurations from our end. However, we must bare in mind that CameraX is library that's an abstraction to Camera2 and other related APIs that prima, so it can't really alter the hardware limitations of any device as such (or at least has it's own limits of doing so). 

Specifically speaking of Xiaomi Mi 9T Pro, it seems to only support 1080p (Full HD) at 30 FPS at best, so I don't really think our camera app would be able to reliably capture videos at a higher FPS than that (for example, at 60 FPS) unless the CameraX team somehow comes up with a workaround (which is however highly unlikely considering the ton of workload and feature requests they already have).

We'll add an update to this thread as soon as CameraX includes an option to alter the frame rate/resolution and we include it in our camera app.

Thanks a lot for trying out our camera app @Donkey-Doug!

MHShetty:
> What about resolution? My phone (Xiaomi Mi 9T Pro) only films in 1080p @ 30fps.

We might need to wait until the CameraX team includes this in their public schedule or any of their future release to be able to actually alter these configurations from our end. However, we must bare in mind that CameraX is library that's an abstraction to Camera2 and other related APIs that prima, so it can't really alter the hardware limitations of any device as such (or at least has it's own limits of doing so). 

[...]

PR #397, opened 2023-11-30
Add 60 fps toggle to Camera app

shur-complement:
Hi! If interested, I've gone and added a "60fps" toggle to the camera app. I'm submitting this PR to see if there's interest in this feature, or something _like_ it. The purpose is to enable higher frame rate video capture than the current default. Let me know your thoughts!

This change specifically adds a toggle to the Camera app allowing the user to record video at a higher
frame rate ( 60fps ). The default is unchanged, and this setting is _disabled_  by default.

It's possible that users might want a bit more flexibility, so they could set the frame rate in the "other settings" page, rather than in the inline menu, but I felt this was acceptable as a first attempt.

Related to https://github.com/GrapheneOS/Camera/issues/51

Screenshot:

screenshot

### DIFF
--- a/app/src/main/java/app/grapheneos/camera/CamConfig.kt
+++ b/app/src/main/java/app/grapheneos/camera/CamConfig.kt
@@ -10,6 +10,7 @@ import android.net.Uri
 import android.os.Build
 import android.provider.MediaStore
 import android.util.Log
+import android.util.Range
 import android.util.Size
 import android.view.MotionEvent
 import android.view.View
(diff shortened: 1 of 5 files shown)


thestinger:
It probably needs to detect the available frame rates and provide a choice rather than 1 toggle for this.

shur-complement:
@thestinger I see, thank you for the feedback. I acknowledge that a toggle is not the best option. As far as UX, what is your preference?

One possibility is to show a screen with supported resolutions and framerates, like so:

image

Another could be the way [OpenCamera](https://opencamera.org.uk/) does it, where they show:

image

thestinger:
I'm not exactly sure. Not all frame rates can be used with all resolutions.

erkexzcx:
Missing higher FPS (e.g. 60) is the only reason of why I am not using GrapheneOS Camera app... :( Would love to have 60fps.

shur-complement:
This branch does work if you build + install using android sdk, at least on Pixel devices. I will consider a more general solution. Until then closing this PR.

MHShetty:
The feature would be better implemented as a dropdown to dynamically show different frame rates supported by different resolution

PR #418, opened 2024-02-16
rewrite blur bitmap implementation

empratyush:
it uses RenderEffect for android 12+ and RenderScript with ScriptIntrinsicBlur for older version.


empratyush:
Tested on Pixel 7 pro and Mi A3 (Android 11).

PR #420, opened 2024-02-20
add support for dynamically changing white balance

empratyush:


empratyush:
Tested on Pixel 7 pro.

empratyush:
ui need to redone.

PR #438, opened 2024-06-10
Use CameraX's recordedDuration in status event for video recorder's timer

MHShetty:
This would help us in displaying a more accurate timer while the video recording takes place, for example in cases when the user pauses the recording or when an unexpected occasional delay occurs before the recording actually starts


app/src/main/java/app/grapheneos/camera/capturer/VideoCapturer.kt
```diff
@@ -52,48 +52,33 @@ class VideoCapturer(private val mActivity: MainActivity) {
...
     private val handler = Handler(Looper.getMainLooper())
-    private var elapsedSeconds: Int = 0
-    private val runnable = Runnable {
-        ++elapsedSeconds
-        val secs = padTo2(elapsedSeconds % 60)
-        val min = padTo2(elapsedSeconds / 60 % 60)
-        val hours = padTo2(elapsedSeconds / 3600)
+
+    private fun padTo2(time: Long): String {
+        return String.format("%1$" + 2 + "s", time).replace(' ', '0')
```
muhomorr, line 67: `String.format(Locale.ROOT, "%02d", time)` is a cleaner way to achieve this.

app/src/main/java/app/grapheneos/camera/capturer/VideoCapturer.kt
```diff
@@ -52,48 +52,33 @@ class VideoCapturer(private val mActivity: MainActivity) {
...
+
+    private fun updateTimerTime(timeInNanos: Long) {
+        val timeInSec = timeInNanos / (1000 * 1000 * 1000)
+        val secs = padTo2(timeInSec % 60)
+        val min = padTo2(timeInSec / 60 % 60)
+        val hours = padTo2(timeInSec / 3600)
         val timerText: String = if (hours == "00") {
             "$min:$secs"
         } else {
             "$hours:$min:$secs"
```
muhomorr, line 78: `String.format(Locale.ROOT, "%02d:%02d:%02d", hours, min, sec)` would be more understandable

muhomorr (changes requested):

muhomorr:
This change occasionally causes video recording duration to be stuck at "0:00", not clear why. Saving the recording doesn't work produces the "this recording is too short to be saved" error.

app/src/main/java/app/grapheneos/camera/capturer/VideoCapturer.kt
MHShetty: Yes, it would be much better to use `String.format` over manually calculating this. Have fixed this in the latest changes pushed to the branch

MHShetty:
> This change occasionally causes video recording duration to be stuck at "0:00", not clear why. Saving the recording doesn't work produces the "this recording is too short to be saved" error.

Yes actually this happens because the recording does not directly start as soon as we call `PendingRecording.start` but rather a short while after that. This PR was made to address the recorded time not being in sync with the timer displayed by the app. 

I have pushed some changes to only animate the UI when the recording starts, while the sound plays as the way it used to before `PendingRecording.start` is called, to show a more accurate time state.

If the recording is less than a second or half a second, it would most likely lead to an error like that. However, if it still happens after the latest changes (i.e. animating the UI and timer after CameraX notifies us about the recording starting), it could be a bug that might need to be reported

muhomorr:
Found a bug with latest changes, not sure if it's the same as the previous bug:
- clear storage of Camera app to reset its permissions
- open Camera, grant it camera pemission
- switch to video mode, tap "start recording" button
- allow microphone permission
- observe that recording appears to have been started, but the counter is not counting. Tapping "finish recording" button shows "recording is too short" message

MHShetty:
If the 00:00 red indicator shows up and then doesn't update later, this could be a bug of CameraX. It implies that `VideoRecordEvent.Start` event was called but not the `VideoRecordEvent.Status` which could mean that the recorder isn't recording even after starting. Could you please share the logs specific to the app while reproducing this issue?

For me, it's showing the counter as expected on first launch and attempt to record the video but end up with the same "recording is too short" error. The subsequent recording attempts seem to work as expected even for durations shorter than a second although.

MHShetty:
In the first attempt of recording the video we have,
```
2024-06-27 13:56:55.480 19719-19719 Recorder                app.grapheneos.camera.dev            D  Transitioning Recorder internal state: CONFIGURING --> PENDING_RECORDING
2024-06-27 13:57:04.114 19719-19719 Recorder                app.grapheneos.camera.dev            D  Transitioning Recorder internal state: PENDING_RECORDING --> CONFIGURING
```

This stays almost forever after that, until we attempt to stop recording which then just logs
```
> 2024-06-27 13:57:04.117 19719-19719 Recorder                app.grapheneos.camera.dev            D  Sending VideoRecordEvent Finalize [error: ERROR_NO_VALID_DATA]
```

In the second attempt, however,
```
2024-06-27 13:57:17.181 19719-19719 Recorder                app.grapheneos.camera.dev            D  Transitioning Recorder internal state: IDLING --> PENDING_RECORDING
2024-06-27 13:57:17.183 19719-20009 Recorder                app.grapheneos.camera.dev            D  Transitioning Recorder internal state: PENDING_RECORDING --> RECORDING
```
(rest of the logs as expected and the recording works)

The main difference seems to be between the transition in the internal state (between the logs of the two case), one transitions to CONFIGURING state whereas the other transitions to RECORDING state directly

MHShetty:
This happens specifically when the recording starts right after the recording permission is given, possibly related to the different flows that are taken when audio permission is not already present and when it is

MHShetty:
The bug occurs because `startRecording` is called on the old instance of videoCapturer while `onResume` that gets called after the permission granted callback is completed re-initializes the camera

MHShetty:
Hey @muhomorr,

Could you please test the latest revision of this branch for this issue?

Thanks a lot for your time and help!

Issue #373, opened 2023-07-31
QR scan area shows as black square


nonremittal:
Xiaomi Mi A2 running Android 10 and the Camera app v. 63.

The active scanning area shows up as an opaque black box:
![scanner_screenshot](https://github.com/GrapheneOS/Camera/assets/58081816/df16e023-c464-4fa6-bfcf-c1c229301102)

Scanning itself works fine. Haven't experienced this in any other apps with similar designs (e.g. Firefox for Android's QR scanner appears as intended).

Duggi-Niharika-Sai-06:
Hello,

I came across this issue labeled as  in your repository, and I'm very interested in contributing to it as part of a school project.

I believe working on this issue will not only help me gain practical experience in, but also contribute positively to your project. I've reviewed the contribution guidelines for your repository and am ready to adhere to them.

Could you please let me know if it's possible for me to work on this issue? If there are any specific requirements or starting points you'd like me to know, I'd be glad to hear them.

Thank you for considering my request. I'm looking forward to the opportunity to contribute to your project.

Best regards,
Neethu Duggi

thestinger:
It looks like a bug with the device and it's not clear what can be done about that.

MHShetty:
Hi @nonremittal,

Are you currently active to help us out with resolving this issue?

MHShetty:
Hi @nonremittal,

Are you currently active and willing to help us out with testing a solution for this issue?

Thanks a lot for reporting this issue and trying out our app!

PR #439, opened 2024-06-12
Use blendMode over xfermode

MHShetty:
This PR aims to support QR for devices with hardware acceleration enabled

(Needs testing on a device where the issue #373 is currently being faced on the main branch)

### DIFF
--- a/app/src/main/java/app/grapheneos/camera/ui/QROverlay.kt
+++ b/app/src/main/java/app/grapheneos/camera/ui/QROverlay.kt
@@ -2,11 +2,10 @@ package app.grapheneos.camera.ui
 
 import android.content.Context
 import android.content.res.Resources
+import android.graphics.BlendMode
 import android.graphics.Canvas
 import android.graphics.Color
 import android.graphics.Paint
-import android.graphics.PorterDuff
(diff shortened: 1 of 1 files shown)


muhomorr:
Why do you think the linked issue is due to hardware acceleration? Which kind of hardware acceleration are you referring to?

MHShetty:
> Why do you think the linked issue is due to hardware acceleration? Which kind of hardware acceleration are you referring to?

I wasn't able to find a reference in the official docs directly pointing to `PorterDuff.Mode.CLEAR` regarding this issue, however there are multiple threads on the internet regarding this issue pointing out to the same thing, including this comment made by an official Google ID (with another fix to this issue) - https://issuetracker.google.com/issues/179644462#comment3

> I think it refers to the layer acceleration.

Yes, we're referring to hardware layer rendering mode that devices could have enabled for smoother and faster rendering which seems to be causing `PorterDuff.Mode.CLEAR` xferMode to not work as expected

However, we might still want to test if this fixes the issue on any device was previously facing to confirm that it helps fix the QR mode on devices where hardware acceleration is enabled by default.

Issue #370, opened 2023-07-15
Changing screen orientation in gallery resets current image to first one


mb720:
Hi and thanks a lot for GrapheneOS!

When I'm in the image gallery and rotate the screen, e.g. from portrait to landscape, the gallery resets to show the first image in the gallery.

Then I have to swipe left to get back to the image I wanted to see.

I'm using a Pixel 6 Pro.

The camera app is:

>version 63
>app.grapheneos.camera
>versionCode 63
>targetSdk 33
>minSdk 29

I tried working around this by changing the gallery app to some different app, but according to the discussion in [this feature request](https://github.com/GrapheneOS/Camera/issues/111), that's not possible.

empratyush:
>When I'm in the image gallery and rotate the screen, e.g. from portrait to landscape, the gallery resets to show the first image in the gallery.



Unable to reproduce this. Is this about built-in gallery app in GrapheneOs?

mb720:
The bug still happens on my device.

>Is this about built-in gallery app in GrapheneOs?

Yes, the one that opens when you click here inside the camera app:

<img src="https://i.imgur.com/VOE4OrN.png" alt="Camera app screenshot" width="30%"/>

The camera app is:

    version 63
    app.grapheneos.camera
    versionCode 63
    targetSdk 33
    minSdk 29

tsxv478:
Can confirm that this happens when swiping between full screen images in both grid and filmstrip views, except with grid view it's not the first image in the gallery, but rather the one you last opened. To reproduce: open any image from the grid view, swipe to another one without exiting full screen view and then rotate the screen.

mb720:
I'm not sure if this was fixed in version 66, which I'm using now, but the issue doesn't occur at the moment. Thanks for fixing this!

@tsxv478: Note that the issue I referred to happened in the camera app's gallery, not the app named "Gallery" that supports filmstrip view and grid view.

MHShetty:
The issue has been fixed in the PR mentioned above #440

mb720:
I'm using the camera app version 74 and this seems to be fixed. Thanks a lot 👍

PR #440, opened 2024-06-12
Fix in-app gallery resetting to the first image on orientation change

MHShetty:
Changing the orientation in the in-app gallery currently causes the entire gallery activity to reset

This pull request would help resolve the issue described at #370


muhomorr:
Activity is recreated after almost all configuration changes, not just after rotation. Proper solution for this is to use a ViewModel or to save activity state in onSaveInstanceState(). 

The proposed solution doesn't work in my testing even for orientation changes, not clear why.

muhomorr:
Special handling of lastCapturedItem should be skipped when lastViewedMediaPosition is not 0, otherwise lastCapturedItem will be loaded unnecessarily before item at lastViewedMediaPosition.

MHShetty:
> Special handling of lastCapturedItem should be skipped when lastViewedMediaPosition is not 0, otherwise decoding of item at lastViewedMediaPosition will be delayed until decoding of lastCapturedItem is complete.

Yes, loading the first item would be un-necessary in this case. Thanks a lot for reviewing this PR!

muhomorr:
Item position is not guaranteed to be stable after the item list is re-fetched, missed that initially. A stable item identifier should be saved in `onSaveInstanceState`, not its position.

Issue #323, opened 2022-11-08
Allow mute mic in video.


balika011:
Some camera apps allows to mute the mic in video recording.
I use that a lot and miss that feature.

thestinger:
There's an include audio feature already. Do you mean a dynamic toggle?

MHShetty:
The dynamic toggling of audio during a video recording can now be implemented as it's supported by the CameraX library as of version `1.3.0-alpha06`

PR #441, opened 2024-06-12
Add support to dynamically mute mic in-recording

MHShetty:
Enables users to dynamically include/exclude audio while recording a video 

(Solves issue #323)

### DIFF
--- a/app/src/main/java/app/grapheneos/camera/capturer/VideoCapturer.kt
+++ b/app/src/main/java/app/grapheneos/camera/capturer/VideoCapturer.kt
@@ -352,6 +352,14 @@ class VideoCapturer(private val mActivity: MainActivity) {
         mActivity.forceUpdateOrientationSensor()
     }
 
+    fun muteRecording() {
+        recording?.mute(true)
+    }
+
+    fun unmuteRecording() {
+        recording?.mute(false)
+    }
+
     fun stopRecording() {
         recording?.stop()
         recording?.close()
--- a/app/src/main/java/app/grapheneos/camera/capturer/VideoCapturer.kt
+++ b/app/src/main/java/app/grapheneos/camera/capturer/VideoCapturer.kt
@@ -283,7 +283,7 @@ class VideoCapturer(private val mActivity: MainActivity) {
 
         animator.start()
 
-        mActivity.settingsDialog.includeAudioToggle.isEnabled = false
+//        mActivity.settingsDialog.includeAudioToggle.isEnabled = false
         mActivity.settingsDialog.videoQualitySpinner.isEnabled = false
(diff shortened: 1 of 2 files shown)


app/src/main/java/app/grapheneos/camera/ui/SettingsDialog.kt
```diff
@@ -324,11 +327,33 @@ class SettingsDialog(val mActivity: MainActivity) :
 
         includeAudioToggle = binding.includeAudioSwitch
         includeAudioToggle.setOnClickListener {
+            if (mActivity.videoCapturer.isRecording) {
+                if (ActivityCompat.checkSelfPermission(mActivity, Manifest.permission.RECORD_AUDIO)
+                    != PackageManager.PERMISSION_GRANTED) {
+
+                    // Inform the user why enabling this option isn't possible
+                    mActivity.showMessage("Unable to request for audio permission in between a recording")
```
muhomorr, line 335: UI strings should be specified via string resources

app/src/main/java/app/grapheneos/camera/capturer/VideoCapturer.kt
```diff
@@ -283,7 +283,7 @@ class VideoCapturer(private val mActivity: MainActivity) {
 
         animator.start()
 
-        mActivity.settingsDialog.includeAudioToggle.isEnabled = false
+//        mActivity.settingsDialog.includeAudioToggle.isEnabled = false
```
muhomorr, line 286: Why comment out this code instead of removing it?

muhomorr (changes requested):

app/src/main/java/app/grapheneos/camera/ui/SettingsDialog.kt
MHShetty: Have pushed a commit to the branch to extract the string into strings.xml

app/src/main/res/values/strings.xml
```diff
@@ -185,4 +185,5 @@
 
     <string name="zsl_setting_title">Use ZSL in Latency mode</string>
     <string name="zsl_setting_desc">Uses Zero Shutter Lag (ZSL) in Latency mode for faster capture. Certain devices may have a buggy implementation for this.</string>
+    <string name="unable_to_request_for_audio_permission_in_between_a_recording">audio_permission_failed_in_recording</string>
```
muhomorr, line 188: This string value seems to be a mistake

muhomorr (changes requested):

app/src/main/res/values/strings.xml
MHShetty: Ah. Thanks for informing!

Issue #391, opened 2023-10-26
EIS not fully working on Pixel 6


Silent-Hunter:
EIS works fine when not recording. I can tilt the phone slightly and the image in the viewfinder is stable. But as soon as I start recording, it's as if the EIS is off. It's been like this for over a year.

GrapheneOS build number is UP1A.231005.007.2023102300

Graphene Camera is version 63

soupslurpr:
It seems to be intentional

Gets disabled before recording starts here https://github.com/GrapheneOS/Camera/blob/53efbad43aaaeb3a8996c6c77da9d84b2de22dda/app/src/main/java/app/grapheneos/camera/capturer/VideoCapturer.kt#L288

Gets re-enabled when recording stops here
https://github.com/GrapheneOS/Camera/blob/53efbad43aaaeb3a8996c6c77da9d84b2de22dda/app/src/main/java/app/grapheneos/camera/capturer/VideoCapturer.kt#L326

Silent-Hunter:
What on Earth? Why? That seems odd.

thestinger:
@soupslurpr That code prevents changing the state of the EIS toggle after starting the video. It doesn't disable EIS.

soupslurpr:
Ah, sorry for misunderstanding that. Is it a bug then that EIS doesn't seem to work when starting the recording? I noticed this too on my Pixel 7

ASerbinski:
I can confirm that reports look valid. Move the EIS switch to "on" and the field of view narrows to provide the feature's required buffer, and shaking the phone around a bit, the image appears to be stabilized. As soon as recording is enabled though, the video no longer appears to be stabilized, but still has the narrowed field of view.

My first thought is that it could be connected to some kind of logic that may disable EIS if the camera has OIS capability. Of note, Pixel 6 does have OIS on the main lens, but if I remember correctly, not on the wide angle or telephoto (pro model) lenses.

MHShetty:
Hey everyone,

Could someone please help us out with testing a fix for this issue?

Silent-Hunter:
How would one do that?

MHShetty:
> How would one do that?

Hey @Silent-Hunter, are you present on Element?

soupslurpr:
I can test too :)

Silent-Hunter:
> Hey @Silent-Hunter, are you present on Element?
No, sorry.

MHShetty:
> > Hey @Silent-Hunter, are you present on Element?
[...]

PR #444, opened 2024-06-17
Use CameraX's official APIs for Video Stabilization

MHShetty:
This PR replaces the currently used Camera2Interop API for EIS with CameraX's official APIs for video stabilization

(It could possibly solve the bug being described in Issue #391; however, we would need testing and confirmation from the user to ensure there isn't any device-specific issue with EIS to be reported further)

### DIFF
--- a/app/src/main/java/app/grapheneos/camera/ui/SettingsDialog.kt
+++ b/app/src/main/java/app/grapheneos/camera/ui/SettingsDialog.kt
@@ -332,7 +332,7 @@ class SettingsDialog(val mActivity: MainActivity) :
                     != PackageManager.PERMISSION_GRANTED) {
 
                     // Inform the user why enabling this option isn't possible
-                    mActivity.showMessage("Unable to request for audio permission in between a recording")
+                    mActivity.showMessage(context.getString(R.string.audio_permission_failed_in_recording))
 
                     // Ensure the option is visually off
                     includeAudioToggle.isChecked = false
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -189,4 +189,5 @@
 
     <string name="zsl_setting_title">Use ZSL in Latency mode</string>
     <string name="zsl_setting_desc">Uses Zero Shutter Lag (ZSL) in Latency mode for faster capture. Certain devices may have a buggy implementation for this.</string>
+    <string name="audio_permission_failed_in_recording">Unable to request for audio permission in between a recording</string>
 </resources>


app/src/main/java/app/grapheneos/camera/CamConfig.kt
```diff
@@ -532,6 +531,10 @@ class CamConfig(private val mActivity: MainActivity) {
         camera!!.cameraInfo.isZslSupported
     }
 
+    val isStabilizationSupported : Boolean by lazy {
+        Recorder.getVideoCapabilities(camera!!.cameraInfo).isStabilizationSupported
```
muhomorr, line 535: Is `camera` guaranteed to be non-null at this point?

MHShetty: Thanks a lot for pointing this out @muhomorr!

While this is guaranteed to be true for the normal `CaptureActivity`, it wouldn't have been true when the video mode has been directly invoked for example in case of `android.media.action.VIDEO_CAMERA` intent or similar intents supported by our app.

To fix this issue, we could instead rely on getting the list of `CameraInfo` objects from the `cameraProvider` object present in the `CamConfig` class that is guaranteed to be non-null when `startCamera` method is called and then select the `CameraInfo` object based on its `lensFacing` which would be guaranteed to be available otherwise  since we ensure that before calling `startCamera` in our app.

The `CameraInfo` object that would then be retrieved using the above method could be used to be passed instead to make this robust regardless of the use cases being bind or not.

I have pushed the changes related to the same to the branch. Please do let me know if any further changes are needed. Thanks for your valuable time and help!

app/src/main/java/app/grapheneos/camera/CamConfig.kt
```diff
@@ -894,6 +898,14 @@ class CamConfig(private val mActivity: MainActivity) {
         startCamera(true)
     }
 
+    // Will always return a valid CameraInfo if called within startCamera
+    private fun getCurrentConfigCameraInfo() : CameraInfo? {
+        for (cameraInfo in cameraProvider!!.availableCameraInfos) {
+            if (cameraInfo.lensFacing == lensFacing) return cameraInfo
```
muhomorr, line 904: There can be more than one camera with same lensFacing value, correct? This code doesn't account for that

MHShetty: Our app only supports handling a single camera for each `lensFacing` currently. We could stick to saving info of a single camera for each side (into variables `frontCameraInfo` and `rearCameraInfo`) and use the physical camera ID present in that info to select and display information about the right camera to avoid ambiguity just in case if the order in which availableCameras are returned is uncertain/could change in the future

MHShetty: Our camera app currently only supports using only one camera for each side facing which is chosen by the CameraX library. However, considering the possibility that the way CameraX chooses the camera could differ from our chosen camera info (now/later in case of multiple cameras), we could explicitly choose a cameraInfo for both front and rear facing (if they are available) and ensure that the app only chooses that camera we're showcasing the information for. Support for selection of camera can be dealt with a separate PR as our app currently displays the camera chosen by CameraX by default.

PR #450, opened 2024-07-23
Disable including audio in-recording when it starts disabled

MHShetty:
Temporary workaround until we have a reliable way to start recording with mic mute while allowing the user to unmute the mic later


MHShetty:
To keep track of the status of the issue being workaround here, here's the link to the issue filed at CameraX for the same: https://issuetracker.google.com/issues/354829267

MHShetty:
The CameraX team has mentioned that they'll be overloading the withAudioEnabled method with a **muted** parameter to be to help resolve the issue we're facing at the moment (https://issuetracker.google.com/issues/354829267#comment2)

PR #453, opened 2024-07-25
Separate mute toggle from include audio toggle option

MHShetty:


muhomorr:
Mute/unmute button is too close to the "stop recording" button. It's easy to accidentally tap the wrong button, especially when the device and/or user is not stationary.

MHShetty:
Yes that's true. How about we position the toggle above the pause button as shown in the above screenshot?

PR #468, opened 2024-08-30
Migrate in-app gallery to Jetpack Compose

MHShetty:


app/src/main/java/app/grapheneos/camera/ui/composable/screen/GalleryScreen.kt
```diff
@@ -0,0 +1,349 @@
...
+                                )
+                                return@launch
+                            }
+
+                            val currentItem = getCurrentMediaItem()
+
+                            val editIntent = Intent(Intent.ACTION_EDIT).apply {
+                                setDataAndType(currentItem.uri, currentItem.mimeType())
+                                putExtra(Intent.EXTRA_STREAM, currentItem.uri)
+                                addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION or Intent.FLAG_GRANT_WRITE_URI_PERMISSION)
```
muhomorr, line 203: Why is `Intent.FLAG_GRANT_WRITE_URI_PERMISSION` added here? It wasn't used by the previous implementation for this intent.

app/src/main/java/app/grapheneos/camera/ui/composable/screen/GalleryScreen.kt
```diff
@@ -0,0 +1,349 @@
...
+
+        coroutineScope.launch {
+            val items = mediaItemsLoader()
+
+            if (items.isEmpty()) {
+                Toast.makeText(context, R.string.empty_gallery, Toast.LENGTH_LONG).show()
+                onExit()
+            }
+
+            capturedItems.clear()
```
muhomorr, line 127: Missing synchronization, and ordering between these 2 coroutines doesn't look right. There's no guarantee about which will finish first.

app/src/main/java/app/grapheneos/camera/ui/composable/components/TopBarActions.kt
```diff
@@ -0,0 +1,172 @@
...
+                            )
+                        }
+                    }
+
+                    DropdownMenu(
+                        expanded = dropDownMenuExpanded,
+                        onDismissRequest = {
+                            dropDownMenuExpanded = false
+                        },
+                        modifier = Modifier.width(200.dp)
```
muhomorr, line 137: It's not guaranteed that screen width is over 200 dp, it can be less than that on smaller devices when screen scaling is used.

app/src/main/java/app/grapheneos/camera/ui/composable/screen/GalleryScreen.kt
MHShetty: It was initially added to test whether the image updates after editing it from another app (and fixes around it). This issue still exists due to the limited options present in the library (to re-compose the image) after returning back to the app. I have added a separate toggle for in-place editing in the local commits. Will update this thread as soon as the main issue gets fixed and the changes are pushed

app/src/main/java/app/grapheneos/camera/ui/composable/components/TopBarActions.kt
MHShetty: Ah that makes sense, will update the thread once I push the changes for the same

app/src/main/java/app/grapheneos/camera/ui/composable/screen/GalleryScreen.kt
MHShetty: We have added checks to avoid unexpected behavior irrespective of which finishes first (`latestItem != null && capturedItems.isEmpty()`). We could merge it into a single coroutine callback to keep it simple and avoid the need for synchronization altogether

app/src/main/java/app/grapheneos/camera/ui/composable/screen/GalleryScreen.kt
MHShetty: Have updated the code to ensure things work as expected. A separate option is now present to support editing the original image while the behavior of the original options has been set as before

app/src/main/java/app/grapheneos/camera/ui/composable/components/TopBarActions.kt
MHShetty: We're filling up a certain ratio of the screen width based on the screen in which the top bar is situated

muhomorr:
What's the purpose of new (de)serializers for CapturedItem and Uri? They can be (de)serialized already since they implement the Parcelable interface.

MHShetty:
Both the Kotlin Serializable and the new Parcelable interfaces are needed for using type-safe navigation/custom objects in nav types. Initially, we were passing CapturedItem and List<CapturedItem> to specific routes/screen as parameters, but we later switched to having an activity-level repository to manage the captured items which is a much better and cleaner approach.

The only serializer and parcelable implementation we are currently using in the app is VideoUriSerializer class and the Parcelable implemented along with the VideoUri class itself. I have kept the other classes just in case if we need them later within the camera app or the gallery app. Type-safe navigation isn't very well documented unfortunately, a lot of the limitations it has isn't very clearly mentioned, so having the other serializable and nav type classes as reference would be helpful for future use

muhomorr:
What I'm saying is that Kotlin (de)serializers can reuse the implementation of Parcelable (de)serializers instead of duplicating them.

muhomorr:
A better solution in this case would probably be to not use route objects at all, since they don't work well with Parcelables. There's an option to pass route arguments as a Bundle instead. Bundle supports Parcelables directly.

MHShetty:
Route objects help us to do type-safe navigation by ensuring the right parameters are called (with respect to their type, order and in avoiding null check/assertions), and also improving the code readability while performing navigate calls apart from the compile-time checks it provides. Since we have already written the code for the same, I feel it's better to go ahead with the standard type-safe navigation approach

Issue #473, opened 2024-09-12
False positive selection error (inconsistent)


An-anonymous-coder:
# Description
Occasionally while selecting and deselecting options in the QR Scan tab, the error message "Please ensure that at least one barcode is selected in manual mode". This is inconsistent behavior, but it's consistent enough that if you try the reproducible steps a couple times it should work. 

# Steps
1. Open the camera app
2. Select the "QR SCAN" tab
3. Switch to manual mode by clicking the "X" in the bottom left. Ignore this step if it is already in manual mode. 
4. Deselect all scan options on the top.
5. Open the "More Options" window by tapping on the three dots in the top right.
6. Deselect all options except for a single option (e.g. "CODABAR").
7. Click "OK" in the bottom right.
8. Open the "More Options" window by tapping on the three dots in the top right.
9. Select another option and deselect your first option (e.g. select "CODE_39" and deselect "CODABAR"). It does not matter if you select the second option before or after deselecting the first option (e.g. you can deselect "CODABAR" before selecting "CODE_39" and it will still reproduce the bug).
11. Click "OK" in the bottom right.
12. If no error message pops up, repeat steps 8 through 10.

# Unexpected behavior
Eventually, the error message "Please ensure that at least one barcode is selected in manual mode" should pop up, despite an option being selected. Opening the "More Options" menu by tapping the three dots in the top right shows that now both options have been selected.

# Expected behavior
No error message should appear. Only one option should be selected.

# Video example
In the video, it does not error the first time trying steps 8 through 10 but does error the second time. I have found that it usually does not error on the first try.

https://github.com/user-attachments/assets/ae2d7ea8-db35-4d2f-a6f7-325df1bf771b

MHShetty:
Hi @An-anonymous-coder,

Could you please test this apk for this issue?

https://drive.proton.me/urls/B2K0BTBM24#EmTKQjhs0uBK

MHShetty:
Hi @An-anonymous-coder,

Thanks a lot for testing out the camera app and reporting this issue!

Could you please test the apk present in the following link for this issue?

https://drive.proton.me/urls/B2K0BTBM24#EmTKQjhs0uBK

PR #475, opened 2024-09-12
Fix the way at least one barcode selection is ensured

MHShetty:
Fixes the issue described in issue #473

### DIFF
--- a/app/src/main/java/app/grapheneos/camera/CamConfig.kt
+++ b/app/src/main/java/app/grapheneos/camera/CamConfig.kt
@@ -1481,6 +1481,20 @@ class CamConfig(private val mActivity: MainActivity) {
         // Add OK and Cancel buttons
         builder.setPositiveButton(getString(R.string.ok)) { _, _ ->
 
+            val commonFormatsEnabled = commonFormats.count {
+                allowedFormats.contains(it)
+            }
+
+            if (commonFormatsEnabled == 0) {
+                val otherFormatsEnabled = optionValues.count { it }
+                if (otherFormatsEnabled == 0) {
+                    mActivity.showMessage(
+                        getString(R.string.no_barcode_selected)
+                    )
+                    return@setPositiveButton
+                }
+            }
+
             commonPref.edit {
                 for (index in 0 until optionNames.size) {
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/app/grapheneos/camera/CamConfig.kt
```diff
@@ -1481,6 +1481,20 @@ class CamConfig(private val mActivity: MainActivity) {
         // Add OK and Cancel buttons
         builder.setPositiveButton(getString(R.string.ok)) { _, _ ->
 
+            val commonFormatsEnabled = commonFormats.count {
```
muhomorr, line 1484: `none` would be more efficient than `count` in this case.

app/src/main/java/app/grapheneos/camera/CamConfig.kt
```diff
@@ -1481,6 +1481,20 @@ class CamConfig(private val mActivity: MainActivity) {
...
+                allowedFormats.contains(it)
+            }
+
+            if (commonFormatsEnabled == 0) {
+                val otherFormatsEnabled = optionValues.count { it }
```
muhomorr, line 1489: Same as the above.

app/src/main/java/app/grapheneos/camera/CamConfig.kt
```diff
@@ -1492,20 +1506,15 @@ class CamConfig(private val mActivity: MainActivity) {
...
-                    } else if (format in allowedFormats) {
-                        if (allowedFormats.size == 1) {
-                            mActivity.showMessage(
-                                getString(R.string.no_barcode_selected)
-                            )
-                        } else {
+                        if (format !in allowedFormats)
+                            allowedFormats.add(format)
+                    } else {
+                        if (format in allowedFormats) {
```
muhomorr, line 1512: `format in allowedFormats` check is redundant, `allowedFormats.remove(format)` handles this itself.

muhomorr (changes requested):

app/src/main/java/app/grapheneos/camera/CamConfig.kt
MHShetty: Yes that's true. I have updated this in the latest changes pushed to the branch. Thanks for reviewing!

PR #478, opened 2024-09-24
Gracefully cancel image capture request

MHShetty:
Cancels image request in cases when the camera instance restarts (to cover all possible cases such as opening in app gallery, sharing media by long tapping the preview, settings is opened, modes are switched, the side facing is flipped, or any other such case), and when the app gets paused


app/src/main/java/app/grapheneos/camera/capturer/ImageSaver.kt
```diff
@@ -324,6 +332,7 @@ class ImageSaver(
 
     // implementation of ImageCapture.OnImageCapturedCallback.onError
     override fun onError(exception: ImageCaptureException) {
+        if (isCancelled) return
```
muhomorr, line 335: It's not thread-safe to read value of `isCancelled` in `onError` or in `onCaptureSuccess()`. These callbacks are not executed on the main thread.

muhomorr: Using an AtomicBoolean doesn't resolve this race condition fully. Main thread callback can be scheduled already by the time `isCancelled` is set. Cancellation check should happen on the main thread.

MHShetty: Have updated the implementation to ensure both reading and writing happens on the same main thread. Thanks for looking into this!

app/src/main/java/app/grapheneos/camera/capturer/ImageSaver.kt
```diff
@@ -73,17 +74,33 @@ class ImageSaver(
...
+    private var isCancelled = false
+
+    fun cancelCaptureRequest() {
+        isCancelled = true
+    }
+
     override fun onCaptureSuccess(image: ImageProxy) {
-        mainThreadExecutor.execute(imageCapturer::onCaptureSuccess)
+        mainThreadExecutor.execute mainExecutor@ {
+            if (isCancelled) return@mainExecutor
```
muhomorr, line 85: What's the point of this change? Why prevent already captured image from being saved?

app/src/main/java/app/grapheneos/camera/capturer/ImageSaver.kt
```diff
@@ -73,17 +74,33 @@ class ImageSaver(
...
 
-        try {
-            extractJpegBytes(image)
-        } catch (e: Exception) {
-            handleError(ImageSaverException(Place.IMAGE_EXTRACTION, e))
-            return
+            imageCapturer.onCaptureSuccess()
+
+            imageCaptureCallbackExecutor.execute iccExecutor@ {
```
muhomorr, line 89: Why is this call delayed until after `imageCapturer.onCaptureSuccess()`?

app/src/main/java/app/grapheneos/camera/capturer/ImageCapturer.kt
```diff
@@ -25,11 +25,14 @@ import app.grapheneos.camera.ui.activities.SecureMainActivity
 import app.grapheneos.camera.util.printStackTraceToString
 
 private const val imageFileFormat = ".jpg"
-var isTakingPicture: Boolean = false
 
 class ImageCapturer(val mActivity: MainActivity) {
     val camConfig = mActivity.camConfig
 
+    var isTakingPicture: Boolean = false
```
muhomorr, line 32: isTakingPicture variable is redundant, it's always false when currentImageSaver is null and always true when currentImageSaver is non-null.

muhomorr (changes requested):

app/src/main/java/app/grapheneos/camera/capturer/ImageCapturer.kt
MHShetty: Yes, we could set currentImageSaver to null instead of setting a separate variable. I'll push changes for the same soon

app/src/main/java/app/grapheneos/camera/capturer/ImageSaver.kt
MHShetty: We could re-order the calls here. I'll make the changes for the same soon.

app/src/main/java/app/grapheneos/camera/capturer/ImageSaver.kt
MHShetty: This change was added as a safety check to deal with the app going in background (via onPause) as the remaining execution could get uncertain since the app is already in background if the cancelled flag is set to true

muhomorr: What do you mean by "execution could get uncertain"?

MHShetty: I mean the certainty of the proper completion of the execution

muhomorr: There's never such certaintly. App process can be stopped at any time, including when the app is the top foreground app.

muhomorr: Consider what happens after this change when the Camera app activity gets paused due to another activity being launched on top by the OS. That happens automatically in some cases. If such launch happens right after image capture then the image would get discarded.

MHShetty: That's true actually. I'll revert the changes to continue saving the image once a capture is successful as before

MHShetty: Have pushed the changes. Thanks for reviewing!

PR #508, opened 2025-02-07
Auto finish secure activities when the screen goes off

MHShetty:
This PR would ensure that activities are auto-finished when the screen is off and the user is inactive (or the screen doesn't turn on) for a certain short duration

### DIFF
--- a/app/src/main/java/app/grapheneos/camera/AutoFinishOnSleep.kt
+++ b/app/src/main/java/app/grapheneos/camera/AutoFinishOnSleep.kt
@@ -1,0 +1,53 @@
+package app.grapheneos.camera
+
+import android.app.Activity
+import android.content.BroadcastReceiver
+import android.content.Context
+import android.content.Intent
+import android.content.IntentFilter
+import android.os.Handler
+import android.os.Looper
+import androidx.core.app.ActivityCompat
+
+// Finishes the passed [activity] and the ones present below it in the stack if the screen
+// turns off and the user is inactive for [WAITING_TIME_IN_MS] milliseconds
+class AutoFinishOnSleep(val activity: Activity) {
+
+    companion object {
+        private const val TAG = "AutoFinishOnSleep"
+        private const val WAITING_TIME_IN_MS = 1500L
+
+        private val intentFilter = IntentFilter().apply {
+            addAction(Intent.ACTION_SCREEN_ON)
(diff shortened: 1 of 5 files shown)


muhomorr:
`SecureCaptureActivity` and `QrTile` activity aren't handled. 

I think it's better to implement this behavior via application-level activity callbacks (`Application.registerActivityLifecycleCallbacks()`), which would avoid code duplication and would allow requiring that auto-finish is handled for all activities.

MHShetty:
Hi @muhomorr,

We do not want to have this behavior for all activities of our application but rather only the secure ones where media could have been previously captured while the phone is in secure state and the main user wouldn't want anyone else to be able to view the captured media without having to unlock the phone.

`SecureCaptureActivity` captures a photo on behalf of another app but in secure mode. Finishing it would push the user back to the previous activity which can be undesirable behavior for the use case we're trying to support.

muhomorr:
> We do not want to have this behavior for all activities of our application but rather only the secure ones 

I didn't mean to say that all activities should have auto-finishing behavior. I meant that adding an activity should require deciding up-front whether it has to be auto-finished. Otherwise it's quite likely that auto-finishing behavior will be missed for some future secure mode activity. 

> SecureCaptureActivity captures a photo on behalf of another app but in secure mode. Finishing it would push the user back to the previous activity which can be undesirable behavior for the use case we're trying to support.

Whole activity stack can be closed, not just the top activity. Also, commit title says "Auto finish secure activities", which I'm reading as "auto finish **all** secure activities".

MHShetty:
> I didn't mean to say that all activities should have auto-finishing behavior. [...] Also, commit title says "Auto finish secure activities", which I'm reading as "auto finish all secure activities".

Yes, I'm referring to all activities in the same context, we mainly want to apply this behavior for the activities where media could have been previously captured in secure mode and the main user wouldn't want anyone else to be able to view the captured media without having to unlock the phone (assuming someone else accesses the phone gets unlocked)

> Whole activity stack can be closed, not just the top activity. 

Yes, we are currently finishing all the activities in the stack when the screen turns off for a while.

> I meant that adding an activity should require deciding up-front whether it has to be auto-finished. Otherwise it's quite likely that auto-finishing behavior will be missed for some future secure mode activity.

Yes, that's right, I'll push the changes for the same soon. Thanks for the feedback!

Issue #534, opened 2025-04-12
When recording multiple horizontal videos in succession, the second video is incorrectly oriented vertically.


pHr34kY:
When recording multiple horizontal videos in succession, the second video has incorrect orientation, and is recorded vertically. Sorry for duplicating #506 - I didn't realise until after I created this.

To reproduce:
1. Orient the device horizontally
1. Open the GrapheneOS Camera app
2. Select "Video" mode
3. Press the "record" button 6 times, approximately two seconds apart.

**Expected result:** 3 videos recorded in horizontal (landscape) orientation.
**Actual result:** First video recorded as horizontal, second video recorded as portrait, third video recorded as horizontal.
**Suspected cause:** The "Rotation" tag is missing from the metadata of the second video.

Three videos as they appear in the gallery:

![Image](https://github.com/user-attachments/assets/faf936cd-e795-43c3-a298-eed1c86ab7d0)

The same three videos when copied to a PC (Ubuntu):
![Image](https://github.com/user-attachments/assets/0a9b8deb-1f16-48ca-8c47-4d44c8fbcd73)

I made a video demonstrating the issue. It's an external video of the device to demonstrate the orientation.
https://github.com/user-attachments/assets/25898f13-3f03-4e41-a3cf-b271d66cb557

I have included a mediainfo dump of the first two videos. Note that the first video is 1080x1920 with a 270° rotation (a weird way to do it, but it does display correctly in 16:9 landscape). The second video is 1080x1920 (9:16) with no rotation tag on the file.

### First recording
```
$ mediainfo VID_20250412_202654.mp4
General
Complete name                            : VID_20250412_202654.mp4
Format                                   : MPEG-4
Format profile                           : Base Media / Version 2
Codec ID                                 : mp42 (isom/mp42)
File size                                : 4.56 MiB
Duration                                 : 2 s 339 ms
Overall bit rate                         : 16.3 Mb/s
Frame rate                               : 29.868 FPS
Encoded date                             : 2025-04-12 10:26:57 UTC
Tagged date                              : 2025-04-12 10:26:57 UTC
xyz                                      : [redacted]
com.android.version                      : 15

Video
ID                                       : 1
Format                                   : AVC
Format/Info                              : Advanced Video Codec
Format profile                           : High@L4
Format settings                          : CABAC / 1 Ref Frames
Format settings, CABAC                   : Yes
Format settings, Reference frames        : 1 frame
Codec ID                                 : avc1
[...]

PR #535, opened 2025-04-14
Replace orientation API calls with sensor calculated orientation

MHShetty:
The current APIs we use are deprecated/always return 0 since the orientation is fixed for main activity. The rationale behind using accelerometer to reliably calculate orientation is that gravity constantly has small impacts on the accelerometer causing it to continuously generate new values on a real device.

Indirectly fixes the issue described in #534 by only relying on sensor data to calculate orientation


mio-19:
this change does not compile anymore

PR #539, opened 2025-04-24
Add support for enabling low light boost

MHShetty:
Allows user to use low light boost when supported and enabled


mio-19:
Thank you for the work. I would like to see this feature. Could you please rebase with the latest main branch?

PR #553, opened 2025-06-19
Support beginning a video recording with the microphone muted

drift8797:
Now that the app is using CameraX 1.5, it is possible to set an initial muted state when recording a video. This PR makes it so that the microphone mute icon displays at all times in video mode, unless the Include Audio toggle is disabled.

Unlike Include Audio, this setting does not persist. This makes it useful as a quicker way to disable audio for a one-off recording or to start a recording muted and unmute audio only at a later time.

One downside compared to disabling Include Audio is that an audio track is always included regardless of whether the microphone was ever unmuted (it will be completely silent if muted for the entire recording). For videos meant to not have any audio, this uses additional storage and might display differently in some players, so I don't think this should replace the Include Audio toggle.

### DIFF
--- a/app/src/main/java/app/grapheneos/camera/CamConfig.kt
+++ b/app/src/main/java/app/grapheneos/camera/CamConfig.kt
@@ -1145,6 +1145,12 @@ class CamConfig(private val mActivity: MainActivity) {
                     } else {
                         View.VISIBLE
                     }
+                mActivity.muteToggle.visibility =
+                    if (includeAudio) {
+                        View.VISIBLE
+                    } else {
+                        View.GONE
+                    }
 
                 val videoCaptureBuilder = VideoCapture.Builder(
                     Recorder.Builder()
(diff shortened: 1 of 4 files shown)


MHShetty:
Hi @drift8797,

As you have rightly mentioned, choosing not to include in the video is different from muting the audio while recording. Always showing the microphone when include audio option is enabled would be a good way to support both options, although I'm unsure if we would want to have too many options at the bottom or if there could be a better way to show the mute toggle.

I'll review and test your branch and provide feedback soon.

Thanks a lot for showing your interest towards contributing to our repositories, value and appreciate all the time and efforts you have put in!

MHShetty:
Hi @drift8797,

I have tested the code across multiple modes and it seems to work fine everywhere. There's some minor re-factoring that can be optionally done that I have suggested above, but others it seems all good.

drift8797:
@MHShetty Thanks for taking a look! I don't think I can see your feedback here, could you try resubmitting it? I'd be happy to make the changes.

I wasn't completely sure about the UI decisions here. I was thinking since this button already exists once the recording starts it makes sense to keep it in one place, but I would be open to implementing alternative options.

MHShetty:
Hi @drift8797,

Here's a screenshot of the code changes suggested that could be made: image

MHShetty:
Hi @drift8797,

Ah apologizes for the inconvenience. I misunderstood what you were trying to convey earlier. The changes shouldn't conflict in that case. We can keep the changes as it is.

Accepting the PR might take a while as we're a bit occupied with our Android 16 porting process which might take a while, but will surely look forward to accepting this PR once everything settles there.

Thanks a lot for contributing to our camera app, appreciate and value all the time and efforts you have put in!

drift8797:
@MHShetty No worries at all, thank you for all you and the team do for GrapheneOS!

PR #637, opened 2026-04-21
fix inverted accuracy check in getOptimalLocation()

thomasbuilds:
`Location.getAccuracy()` is a radius in meters, smaller is better but `getOptimalLocation()` used `>` in its accuracy tiebreaker, so when two fixes arrived within `STALE_LOCATION_THRESHOLD` (11 s) the *less* accurate one replaced the more accurate one.

The selected `Location` is attached as EXIF GPS data on captured photos and as video metadata when "Save location" is enabled, so the bug silently degrades geotag quality (e.g. a ~500 m network fix overwriting a ~5 m GNSS fix).

### DIFF
--- a/app/src/main/java/app/grapheneos/camera/App.kt
+++ b/app/src/main/java/app/grapheneos/camera/App.kt
@@ -100,8 +100,9 @@ class App : Application() {
                     optimalLocation = location
                 } else {
                     // Compare their accuracy instead of time if the difference is below
-                    // threshold
-                    if (location.accuracy > optimalLocation.accuracy) {
+                    // threshold. Location.getAccuracy() is a radius in meters, so a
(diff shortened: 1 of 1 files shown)


thestinger:
Isn't this still an issue? The fix just didn't seem fully correct.

thomasbuilds:
It is, but knowing the reason why this wasn’t addressed is helpful. I’ll look into this further and PR a full fix.

thestinger:
@thomasbuilds There's more wrong with it than what you fixed and I wasn't sure exactly how to deal with it. It was pointed out by someone else.

thomasbuilds:
if you want I can submit a full fix - with the other issues found, could you link/tell me the issue?

PR #638, opened 2026-05-01
Only show a snackbar when QR scan result is copied for Android 12 and lower

### LINKED ISSUE #619: Do not show confirmation snackbar when copying QR scan result
When using the dedicated button to copy the scan result of a QR scan, a confirmation snackbar is shown that says "Copied text to clipboard".

Have you considered not showing this snackbar?

It seems not showing it and instead relying on the Android 13 and higher default UI is "strongly recommend": https://developer.android.com/develop/ui/views/touch-and-input/copy-paste#duplicate-notifications

Camera version 90 on GrapheneOS 2026012101

mrxx0:
To avoid duplicate notifications and to follow android guidelines for the copy to clipboard this PR disable the snackbar when user copy the QR scan result on Android 13 and upper and keep it for Android 12 and lower.

Closes #619

### DIFF
--- a/.github/workflows/validate-gradle-wrapper.yml
+++ b/.github/workflows/validate-gradle-wrapper.yml
@@ -8,4 +8,4 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
-      - uses: gradle/actions/wrapper-validation@v5
+      - uses: gradle/actions/wrapper-validation@v6


muhomorr:
The system clipboard preview isn't shown by the OS in secondary users.

mrxx0:
Closed by mistake, reopened at https://github.com/GrapheneOS/Camera/pull/639

> The system clipboard preview isn't shown by the OS in secondary users.

Thank you for the review, I fixed it

muhomorr:
Secondary users aren't the same thing as non-system users. Profiles of the main user are not affected by this issue. 

I think it would be better to add a feedback for the "copy action" on top of the copy button instead.

mrxx0:
Screenshot_20260504-175904
Screenshot_20260504-175907


What do you think of an UI feedback like this ?

Screenshot_20260504-191330

Screenshot_20260504-191334

In case of extremely long url

muhomorr:
Looks fine, but the color should be dynamic.

mrxx0:
Updated, https://github.com/GrapheneOS/Camera/pull/639

PR #639, opened 2026-05-04
When copying to clipboard qr scan result show an UI feedback

### LINKED ISSUE #619: Do not show confirmation snackbar when copying QR scan result

mrxx0:
To avoid duplicate notifications and to follow android guidelines for the copy to clipboard this PR disable the snackbar when user copy the QR scan result on Android 13 and upper and keep it for Android 12 and lower.

Closes https://github.com/GrapheneOS/Camera/issues/619

### DIFF
--- a/app/src/main/java/app/grapheneos/camera/ui/activities/MainActivity.kt
+++ b/app/src/main/java/app/grapheneos/camera/ui/activities/MainActivity.kt
@@ -1198,17 +1198,29 @@ open class MainActivity : AppCompatActivity(),
             })
 
             val ctc: ImageButton = dialogBinding.copyQrText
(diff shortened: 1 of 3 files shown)


muhomorr:
See https://github.com/GrapheneOS/Camera/pull/638#issuecomment-4370788474

Issue #530, opened 2025-03-31
avoid preferring overly stale locations


thestinger:
Our code currently checks all location services and uses the one with the best accuracy without taking into account that it could be quite stale. Using every location provider is also strange. We should probably just use the OS fused provider, although that might not be universally available. We're likely currently iterating over gps, network and fused which is strange, and the way we're choosing a location isn't right due to ignoring time.

PR #662, opened 2026-07-09
fix geotagging location selection, providers, lifecycle and staleness

thomasbuilds:
Follow-up to #637 and the remaining part of #530. The one-line fix in #637 was incomplete on its own: with the comparison direction corrected, the one-sided staleness check still lets a significantly older but accurate fix win whenever it appears after a fresher one in a candidate list (last known location scans, batched updates), and the wall-clock time comparison it relies on isn't reliable across providers in the first place.

1. **Pick the most accurate recent location for geotagging instead of the least accurate one**: the accuracy tiebreaker was inverted (smaller `getAccuracy()` is better), the staleness check was one-sided (a significantly *older* candidate fell through to the accuracy comparison, making the result order-dependent), and ages were compared via wall-clock `Location.getTime()`, which the platform documentation says "should not be used to order or compare locations". The selection is now: take the newest fix (by `elapsedRealtimeNanos`), and among fixes not significantly older than it, pick the most accurate; missing accuracy ranks last, ties go to the newer fix. It lives in its own file as a plain top-level `getOptimalLocation(locations)` with instrumented tests covering the ordering, staleness, missing-accuracy and threshold-boundary cases. This commit also stops discarding the stored fix: both the batched update callback and the last known location scan now merge it into the selection instead of replacing or skipping it.

2. **Request location from the fused provider alone instead of every provider at once**: previously updates were requested from every provider at once (gps, network, fused, passive), duplicating what fused already combines and forcing the GNSS engine on. The system-app gate from #531 is replaced with `hasProvider(FUSED_PROVIDER)`: availability is the OS's contract, not the caller's privilege. Falls back to gps, network and passive before S, or on the off chance fused is absent on a newer release.

3. **Stop location updates while the activity is paused**: the listener used to stay registered for the process lifetime; now it follows onResume/onPause, and disabling geotagging clears the stored fix.

4. **Stop attaching stale locations to captures**: `getLocation()` returns null for fixes older than 15 minutes, and drops the stored fix so it can't be reused, instead of silently writing an arbitrarily old position into EXIF; the existing location-unavailable message covers that case. `minDistance` goes from 10 m to 0 so stationary users keep receiving fixes (otherwise a still-valid fix ages past any bound).

### DIFF
--- a/app/src/main/java/app/grapheneos/camera/App.kt
+++ b/app/src/main/java/app/grapheneos/camera/App.kt
@@ -12,7 +12,6 @@ import android.view.WindowManager
 import androidx.annotation.RequiresPermission
 import androidx.appcompat.app.AppCompatActivity
 import app.grapheneos.camera.capturer.deleteStalePendingRecordings
-import app.grapheneos.camera.ktx.isSystemApp
 import app.grapheneos.camera.ui.activities.MainActivity
 import com.google.android.material.color.DynamicColors
 import kotlin.concurrent.thread
@@ -98,17 +97,14 @@ class App : Application() {
         }
         isLocationFetchInProgress = true
 
-        val providers = if (applicationInfo.isSystemApp() && Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
-            listOf<String>(LocationManager.FUSED_PROVIDER)
-        } else {
-            locationManager.allProviders
-        }
+        val providers = locationProviders
+
         val lastKnownLocations = providers.map {
             locationManager.getLastKnownLocation(it)
         }
         location = getOptimalLocation(lastKnownLocations + location)
 
-        locationManager.allProviders.forEach { provider ->
+        providers.forEach { provider ->
             locationManager.requestLocationUpdates(
                 provider,
                 2000,
@@ -117,6 +113,20 @@ class App : Application() {
             )
         }
     }
+
+    private val locationProviders: List<String>
+        get() {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S &&
+                locationManager.hasProvider(LocationManager.FUSED_PROVIDER)
+            ) {
+                return listOf(LocationManager.FUSED_PROVIDER)
+            }
+            return locationManager.allProviders.filter {
+                it == LocationManager.GPS_PROVIDER ||
+                    it == LocationManager.NETWORK_PROVIDER ||
+                    it == LocationManager.PASSIVE_PROVIDER
+            }
+        }
 
     fun dropLocationUpdates() {
         isLocationFetchInProgress = false
--- a/app/src/main/java/app/grapheneos/camera/App.kt
+++ b/app/src/main/java/app/grapheneos/camera/App.kt
@@ -19,10 +19,6 @@ import kotlin.concurrent.thread
 
 class App : Application() {
 
-    companion object {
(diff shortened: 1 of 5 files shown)


app/src/main/java/app/grapheneos/camera/App.kt
```diff
@@ -82,32 +85,25 @@ class App : Application() {
...
-                    // Compare their accuracy instead of time if the difference is below
-                    // threshold
-                    if (location.accuracy > optimalLocation.accuracy) {
-                        optimalLocation = location
-                    }
-                }
-            }
-        }
-        return optimalLocation
+    private fun List<Location?>.getOptimalLocation(): Location? {
```
RankoR: This should be properly covered by tests, including edge cases. To do it, would be better to extract it to a top-level function (ideally - in a separate file).

Also, in general we should avoid extension functions unless they are purely private. `getOptimalLocation(locations: List<Location?>): Location?` would be better.

app/src/main/java/app/grapheneos/camera/App.kt
```diff
@@ -126,37 +122,60 @@ class App : Application() {
...
+                0f,
                 locationListener
             )
         }
     }
 
+    // The fused provider combines the underlying providers itself, so use it
+    // alone when it's available (GrapheneOS implements it in the OS; elsewhere
+    // it may be backed by Play services)
+    private val locationProviders: List<String>
```
RankoR, line 124: Pre-S should also use `LocationManager.PASSIVE_PROVIDER`

app/src/main/java/app/grapheneos/camera/App.kt
```diff
@@ -126,37 +122,60 @@ class App : Application() {
...
+        dropLocationUpdates()
+        location = null
+    }
+
+    fun getLocation(): Location? {
+        val location = this.location ?: return null
+        val ageMs = TimeUnit.NANOSECONDS.toMillis(
+            SystemClock.elapsedRealtimeNanos() - location.elapsedRealtimeNanos
+        )
+        return if (ageMs > MAX_LOCATION_AGE) null else location
```
RankoR: `this.location = null` is needed here in case if `ageMs > MAX_LOCATION_AGE`

app/src/main/java/app/grapheneos/camera/App.kt
```diff
@@ -82,32 +85,25 @@ class App : Application() {
...
+        // Among the locations that aren't significantly older than the newest
+        // one, pick the most accurate. getAccuracy() is a radius in meters, so
+        // a smaller value is more accurate; a location without accuracy
+        // information is considered least accurate, and ties go to the newer one
+        return locations.filter {
+            val ageDifferenceMs =
+                TimeUnit.NANOSECONDS.toMillis(newest.elapsedRealtimeNanos - it.elapsedRealtimeNanos)
+            ageDifferenceMs <= STALE_LOCATION_THRESHOLD
+        }.minWithOrNull(compareBy(
+            { if (it.hasAccuracy()) it.accuracy else Float.MAX_VALUE },
```
RankoR: Hard to read, chaining like `{ -it.elapsedRealtimeNanos }.compareBy<Location> { ... }.thenByDescending { it.elapsedRealtimeNanos }` would be better

RankoR (changes requested):
It will also require a rebase after https://github.com/GrapheneOS/Camera/pull/665 merged

RankoR (approved):

Issue #591, opened 2025-12-06
Would you be open to someone implementing a maximum resolution setting?


pieterdd:
I might try to implement a new feature, so I thought I'd pitch it first before investing time.

The resolution at which my camera produces photos that take up more space than I'd like. I often end up resizing them before I upload them to cloud storage. Ideally, the camera app could save photos at a lower resolution so that my phone can automatically sync my photos to cloud storage as-is. While I'm aware of the existing image quality setting, I'd prefer to set the resolution instead.

What I want to try implementing:
- Adding a setting below 'Timer' called 'Max Edge Size' that limits the pixel size of the longest edge to the given value. It's a dropdown menu containing 'Off', '2560px', '1920px' and '1080px'.
- Find some method to either immediately save the photo at a lower resolution or downsize it when needed. If you have a preference for how to do it, let me know. I can imagine you want to avoid a new dependency on an obscure image processing library for this.

Is this a feature you'd be open to including in Camera?

mech01nc01:
I've opened #669 which implements this — a "Max resolution" setting that caps
the long edge (0 = off).

One thing that might address your dependency concern: it needs no image
processing library at all. Instead of resizing after capture, it tells CameraX
to capture at a lower resolution directly (a ResolutionStrategy on the existing
ResolutionSelector), so there's no extra dependency and no post-processing step.

I went with a free pixel value rather than a fixed dropdown (Off/2560/1920/1080),
but that's trivial to change to a dropdown if the maintainers prefer that.

Issue #505, opened 2025-01-26
Camera mode - allow resolution setting


BryanQuigley:
Looking to add a setting to set a resolution lower then the native default.  Taking a look at the code it seems do-able as the QR bit limits the resolution - [here](https://github.com/GrapheneOS/Camera/blob/ecfb1ee7d3dc17e486c72bf5261b657e8c41ac0d/app/src/main/java/app/grapheneos/camera/CamConfig.kt#L1117).

The bigger question (in my mind) is what settings make sense to expose.  A quick test has me coming up with the following for a very simple picture.
2.1MP1080 × 1920 - 192 KB
4.1MP1512 × 2688 - 314.4 KB
8.3MP2160 × 3840 - 534.8 KB
12.5MP3072 × 4080 (native for pixel 8) -  783.2 KB

I've had native resolution for a single picture go up to 10MB,  my goal is to save space mostly by limiting it to a resolution I'll actually use . 

Could be any of the following additional settings
 *  explicit 8MP option 
 * reduced resolution option that is relative to the native resolution, so maybe a 2/3rds and 1/3 option

mech01nc01:
I've opened #669 which implements this — a "Max resolution" setting that caps
the long edge (0 = off).

As you noted, QR scan mode already limits resolution; this uses the same
mechanism (a CameraX ResolutionStrategy on the ResolutionSelector), just for
image capture, so it's no new concept and needs no image-processing dependency —
the photo is captured at the lower resolution directly.

It's an absolute pixel value rather than the relative 2/3–1/3 option you
mentioned, but the intent (smaller files, resolution you actually need) is the same.

PR #669, opened 2026-07-30
Add optional maximum photo resolution cap

mech01nc01:
Adds an optional "Max resolution" setting (More settings) that caps the long
edge of captured photos to a configurable number of pixels. 0 (the default)
keeps the current behaviour of picking the largest available resolution for the
chosen aspect ratio, so nothing changes for existing users unless they opt in.

Implements the maximum-resolution setting requested in #591 and #505.

Motivation: we'd love to use this app in the field (simple, no bloat), but for
documentation photos an 8-12 MP file has no more value to us than a ~1 MP one -
it just costs storage and upload. A resolution cap solves that predictably.

Two things that should make this an easy fit:

- No new dependency and no post-processing. Instead of resizing after capture,
  it tells CameraX to capture at a lower resolution directly - a
  ResolutionStrategy (FALLBACK_RULE_CLOSEST_LOWER_THEN_HIGHER) added to the
  existing ResolutionSelector, next to the aspect ratio strategy. This addresses
  the concern in #591 about pulling in an image processing library.

- It's the same mechanism QR scan mode already uses to limit its resolution
  (noted in #505), just applied to image capture - so it's not a new concept.

Because it caps the long edge (not the aspect ratio), it works independently of
4:3 / 16:9 and across devices, unlike switching to 16:9 which only reduces
resolution on some sensors. On a test device a 1920px cap produced 1920x1440
(4:3) / 1920x1080 (16:9) instead of 4096x3072 / 4096x2304.

I went with a free pixel value  to be most flexible but could also switch it to a fixed dropdown
(Off / 2560 / 1920 / 1080) if you'd prefer that.

### DIFF
--- a/app/src/main/java/app/grapheneos/camera/CamConfig.kt
+++ b/app/src/main/java/app/grapheneos/camera/CamConfig.kt
@@ -129,6 +129,8 @@ class CamConfig(private val mActivity: MainActivity) {
 
             const val WAIT_FOR_FOCUS_LOCK = "wait_for_focus_lock"
 
+            const val MAX_IMAGE_LONG_EDGE = "max_image_long_edge"
+
             const val SELF_TIMER_DURATION = "self_timer_duration"
         }
 
@@ -174,6 +176,9 @@ class CamConfig(private val mActivity: MainActivity) {
             const val SELECT_HIGHEST_RESOLUTION = false
 
             const val WAIT_FOR_FOCUS_LOCK = false
+
+            // 0 = no cap (use the largest resolution for the chosen aspect ratio)
+            const val MAX_IMAGE_LONG_EDGE = 0
 
             const val SELF_TIMER_DURATION = 0
         }
@@ -571,6 +576,21 @@ class CamConfig(private val mActivity: MainActivity) {
         set(value) {
             val editor = commonPref.edit()
             editor.putInt(SettingValues.Key.PHOTO_QUALITY, value)
+            editor.apply()
+        }
+
+    // Cap for the long edge of captured photos, in pixels. 0 = no cap. Keeps documentation
+    // photos small independently of the aspect ratio / sensor.
+    var maxImageLongEdge: Int
+        get() {
+            return commonPref.getInt(
+                SettingValues.Key.MAX_IMAGE_LONG_EDGE,
+                SettingValues.Default.MAX_IMAGE_LONG_EDGE
+            )
+        }
+        set(value) {
+            val editor = commonPref.edit()
+            editor.putInt(SettingValues.Key.MAX_IMAGE_LONG_EDGE, value)
             editor.apply()
         }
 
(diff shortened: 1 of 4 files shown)


BryanQuigley:
Tested it out and it works for me!  Thank you for pursuing this! (I'm from #505)

I've found two issues

1) The option below Max resolution (what you added) is "Use Highest photo resolution".  This has no effect on my Pixel but these options need to be mutual exclusive I think.  

Maybe this means a dropdown makes more sense?  A dropdown also makes it easier to change when a user say wants it back to native for a few specific pictures (and then will change it back).

2) I was able to set 2048 on my device and it worked perfect (that's likely what I want). But I decided to try 1024 and it didn't work- seems my device has a min resolution of 1920

mech01nc01:
> Tested it out and it works for me! Thank you for pursuing this! (I'm from #505)
> 
> I've found two issues
> 
> 1. The option below Max resolution (what you added) is "Use Highest photo resolution".  This has no effect on my Pixel but these options need to be mutual exclusive I think.
> 
> Maybe this means a dropdown makes more sense? A dropdown also makes it easier to change when a user say wants it back to native for a few specific pictures (and then will change it back).
> 
> 2. I was able to set 2048 on my device and it worked perfect (that's likely what I want). But I decided to try 1024 and it didn't work- seems my device has a min resolution of 1920

Thank you so much for testing this, i really appreciate it!

Both fixed: "Max resolution" and "Use highest photo resolution" are now mutually
exclusive, and I've clarified in the description that values below the device's
smallest supported size fall back to that size (which is why 1024 became 1920).

silvaDominic:
Consider adding some before/after UI screens/videos if you're serious about getting this more attention.

Personally, I never expected anything beyond the dropdown style resolution picker that is prevalent in so many other camera apps. Not sure anything beyond or in addition to that will be considered from a UX standpoint.

Also, why the "Max" qualifier? Is it not just the plain resolution that we're picking/setting? Is there a min also? It seems to imply there's at least a "Minimum" resolution setting (or others) that is configurable which wouldn't make sense. Max, min, and everything in between are just general resolutions a user can _pick_.

Similarly, the 0 = Actual max resolution feels unintuitive, no? I wouldn't expect the free-form value to also act as a toggle. I'd expect a toggle "Use max camera resolution" that would disable the free-form field and then that value to be an integer > 0. But as I said, best bet is a dropdown with standard assigned values.

mech01nc01:
> Also, why the "Max" qualifier? Is it not just the plain resolution that we're picking/setting? 

The main reason I went with a numeric field over fixed presets: absolute values like 2560/1920/1080 etc. assume that those sizes are meaningful on every device, but the supported capture sizes vary a lot between sensors. 
With a free value the cap just adapts and CameraX picks the largest supported size at or below whatever you enter, so it works regardless of a device's specific size list, where a fixed list would sometimes be too coarse or mismatched.

Again: the size is applied at capture time via CameraX, nothing is re-scaled after shot and it needs no image-processing library, so it stays dependency-free and lightweight - therefore i stayed with a setting thats compatible with every device.

BryanQuigley:
Tried asking AI to write a drop down version and it kinda works - but it somehow came up with some resolutions that don't work (or maybe things are reversed): https://github.com/GrapheneOS/Camera/compare/main...BryanQuigley:Camera:bq_resolution . Hoping to find some more time to hack it more.

The Max qualifier already exists in the app today.

PR #687, opened 2026-08-05
Initial preparation for rewrite

RankoR:

PR #691, opened 2026-08-11
Refactor preferences

RankoR:
Depends on https://github.com/GrapheneOS/Camera/pull/687
Closes https://github.com/GrapheneOS/Camera/issues/686

`runBlocking` and limited DI are intentional, as rewrite is phased.


app/src/main/java/app/grapheneos/camera/data/settings/repository/SettingsRepository.kt
```diff
@@ -0,0 +1,583 @@
...
+                }
+            }
+        }
+    }
+
+    private fun write(
+        update: (CameraSettings) -> CameraSettings,
+        persist: () -> Unit,
+    ): Flow<Unit> {
+        return unitFlow {
```
sdsantos, line 301: Why a Flow<Unit> instead of a plain `suspend` method?

sdsantos (approved):
Code looks good. 

But I found a bug on the last photo preview. It shows empty (white), and when I tap it, it shows an error message. This persists until I take a photo, then everything works fine. If I open the production Camera app, the last photo preview is fine. Video:

https://github.com/user-attachments/assets/ae251136-00c6-4f9e-8166-58840c2f2804

app/src/main/java/app/grapheneos/camera/data/settings/repository/SettingsRepository.kt
RankoR: I personally prefer `Flow`s because IMO chaining them looks more elegant, and things like catching exceptions / switching dispatchers are easier to read compared to suspend functions

RankoR: Regarding the white preview - it's not a regression, it's pre-existing. Will be fixed during UX/UI refresh.

app/src/main/java/app/grapheneos/camera/data/settings/repository/SettingsRepository.kt
```diff
@@ -0,0 +1,583 @@
...
+internal class SettingsRepositoryImpl @Inject constructor(
+    @SessionPreferences private val commons: SharedPreferences,
+    private val modePreferences: Map<CameraMode, @JvmSuppressWildcards Lazy<SharedPreferences>>,
+) : SettingsRepository {
+
+    private val mutableSettings = MutableStateFlow(
+        run {
+            seedCommonDefaults()
+            readCommonSettings()
+        },
```
m4pl, line 92: The seed and read run in a field initializer, above `slotted`. An `init {}` below the properties makes that safe by construction and drops the `run {}`.

app/src/main/java/app/grapheneos/camera/di/preferences/PreferencesProvidesModule.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    @Provides
+    @ActivityScoped
+    @SessionPreferences
+    fun provideSessionPreferences(
+        @ActivityContext context: Context,
+    ): SharedPreferences {
+        return commonPreferences(
+            context = context,
+            ephemeral = context is SecureActivity,
```
m4pl, line 32: Nothing checks that lockscreen activities actually carry the `SecureActivity` marker. Add a test so it can't be lost?

app/src/main/java/app/grapheneos/camera/data/settings/repository/SettingsRepository.kt
```diff
@@ -0,0 +1,583 @@
...
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.update
+
+interface SettingsRepository {
+
+    val settings: StateFlow<CameraSettings>
+
+    fun refresh()
+
+    fun setAspectRatio(value: Int): Flow<Unit>
```
m4pl, line 32: Every call site unwraps this with `runBlocking { .collect() }`, and forgetting `.collect()` silently drops the write. Would plain `fun setX(value)` work here?

m4pl (approved):

app/src/main/java/app/grapheneos/camera/data/settings/repository/SettingsRepository.kt
RankoR: `runBlocking` is a temp solution to reduce per-PR changes, later it will look better. We should probably add a hard linter rule for that (Android Studio already highlights such things, but it should be a CI failure)

AGENTS.md
```diff
@@ -38,18 +38,20 @@ Each layer splits per feature, and each feature splits by role:
...
       store/            ExtensionAvailabilityStore
     media/
       model/            CapturedItem and friends
-      repository/       CapturedItemStore
-      store/            MediaStoreDataSource, SafDataSource
+      repository/       CapturedItemRepository
+      store/            CapturedItemStore, MediaStoreDataSource, SafDataSource
     location/
       repository/       LocationRepository
   domain/
```
inthewaves, line 57: Should we be expecting this to be updated by agents as it goes in general?

app/src/main/java/app/grapheneos/camera/data/media/store/CapturedItemStore.kt
```diff
@@ -0,0 +1,158 @@
...
+
+        while (trees.size > CapturedItems.MAX_NUMBER_OF_TRACKED_PREVIOUS_SAF_TREES) {
+            // trees.removeLast() requires API level 35 now due to Java adding it
+            trees.removeAt(trees.lastIndex)
+        }
+
+        commons.edit {
+            putPreviousSafTrees(trees, this)
+        }
+    }
```
inthewaves, line 93: Maybe a sequence would be cleaner, e.g. it would only need to take `MAX_NUMBER_OF_TRACKED_PREVIOUS_SAF_TREES` elements instead of having to remove `n - MAX_NUMBER_OF_TRACKED_PREVIOUS_SAF_TREES` elements from the `removeAt` loop.

```kotlin
  val previousTrees = previousSafTrees()
      .asSequence()
      .filterNot { it == treeUri }

  val trees = sequenceOf(treeUri)
      .plus(previousTrees)
      .take(CapturedItems.MAX_NUMBER_OF_TRACKED_PREVIOUS_SAF_TREES)
```

`putPreviousSafTrees` doesn't seem to inherently need a list anyway, since it just does `trees.joinToString`

Although `MAX_NUMBER_OF_TRACKED_PREVIOUS_SAF_TREES` is pretty small anyway, so this is more of a nit

app/src/main/java/app/grapheneos/camera/data/settings/repository/SettingsRepository.kt
inthewaves: I think `Flow`s work when there's a stream of events, but for one-shot updates it puts the onus on the caller to call `.collect()` to ensure the write is done before continuing execution. I'm not sure if there's a use-case for chaining these `set*` calls for a one-shot `Flow<Unit>` vs a sequence of suspend functions

app/src/main/java/app/grapheneos/camera/ui/activities/InAppGallery.kt
```diff
@@ -53,16 +53,22 @@ import app.grapheneos.camera.util.getParcelableExtra
...
 import java.text.SimpleDateFormat
 import java.util.Date
 import java.util.Locale
 import java.util.TimeZone
 import java.util.concurrent.Executors
+import javax.inject.Inject
 import kotlin.properties.Delegates
 
+@AndroidEntryPoint
 class InAppGallery : AppCompatActivity() {
```
inthewaves, line 67: Secure gallery mode is selected through the `isSecureMode` intent extra, but repository DI determines whether to use `LockscreenCapturedItemRepository` by checking whether the activity implements `SecureActivity`. Since `InAppGallery` does not implement that marker, a secure gallery launch receives the ordinary repository.

This does not cause a current behavior bug: the gallery only calls `capturedItems()`, which the lockscreen wrapper delegates unchanged. However, the injected interface also exposes persistent mutation and SAF grant-release operations. If the marker is intended to enforce the lockscreen dependency boundary, consider injecting a narrower read-only gallery repository or using a distinct secure gallery activity. Otherwise, documenting this as an intentional exception would make the split between `isSecureMode` and `SecureActivity` clear

app/src/main/java/app/grapheneos/camera/ui/activities/InAppGallery.kt
inthewaves, line 610: Secure mode appears to have two representations:

- `SecureActivity` is a type-level marker used by dependency injection to select secure/ephemeral repositories
- `INTENT_KEY_SECURE_MODE` is runtime state used by `InAppGallery` to restrict UI behavior, filter media, and manage the lockscreen lifecycle

InAppGallery can therefore have `isSecureMode == true` without being a `SecureActivity`. Its UI behaves securely, but DI still supplies the ordinary `CapturedItemRepository`.

See the other review comment

inthewaves (changes requested):

PR #702, opened 2026-09-03
Replace SharedPreferences with typed data stores

### LINKED ISSUE #686: Move preferences to SettingsRepository

m4pl:
Closes #686


app/src/main/java/app/grapheneos/camera/data/media/store/StoragePrefsMigration.kt
```diff
@@ -0,0 +1,60 @@
...
+import android.content.SharedPreferences
+import androidx.datastore.core.DataMigration
+import app.grapheneos.camera.data.core.store.LEGACY_MEDIA_URIS
+import app.grapheneos.camera.data.core.store.LEGACY_PREVIOUS_SAF_TREES
+import app.grapheneos.camera.data.core.store.LEGACY_STORAGE_KEYS
+import app.grapheneos.camera.data.core.store.LEGACY_STORAGE_LOCATION
+import app.grapheneos.camera.data.core.store.legacyCommonPreferences
+import app.grapheneos.camera.data.core.store.removeLegacyCommonKeys
+
+// The stores migrate the shared file in arbitrary order, so each removes only its own keys.
```
inthewaves, line 13: By migrating to data store (stored in `/data/user/<userid>/app.grapheneos.camera/datastore/`), the SharedPreferences exclusion line in https://github.com/m4pl/Camera/blob/prefs-settings-repo/app/src/main/res/xml/full_backup_content.xml no longer applies. We need to confirm if we should continue excluding preferences

app/src/main/java/app/grapheneos/camera/CamConfig.kt
```diff
@@ -697,45 +553,24 @@ class CamConfig(private val mActivity: MainActivity) {
...
-                COMMON_SHARED_PREFS_NAME,
-                Context.MODE_PRIVATE
-            ).edit {
-                saveLastCapturedItem(item, this)
-            }
-        }
+        runBlocking { capturedItemRepository.saveLastCapturedItem(item) }
 
         lastCapturedItem = item
     }
```
inthewaves, line 575: `updateLastCapturedItem` can now throw `java.io.IOException` from `capturedItemRepository.saveLastCapturedItem`, e.g. full disk errors

app/src/main/java/app/grapheneos/camera/CamConfig.kt
```diff
@@ -406,101 +316,81 @@ class CamConfig(private val mActivity: MainActivity) {
...
+        }
         set(value) {
-            val editor = commonPref.edit()
-            editor.putInt(SettingValues.Key.GRID, GridType.values().indexOf(value))
-            editor.apply()
-
-            field = value
+            settings = runBlocking {
+                settingsRepository.update { it.copy(gridType = value) }
+            }
```
inthewaves, line 337: Feels like this pattern could maybe block main thread, same for `videoQuality` and `flashMode`

inthewaves (changes requested):

app/src/main/java/app/grapheneos/camera/CamConfig.kt
```diff
@@ -307,36 +211,45 @@ class CamConfig(private val mActivity: MainActivity) {
...
+
+    private val preferencesScope = CoroutineScope(Dispatchers.Main.immediate)
 
     var lastCapturedItem: CapturedItem? = null
 
     init {
-        if (mActivity !is SecureActivity) {
-            CapturedItems.init(mActivity, this)
-            fetchLastCapturedItemFromSharedPrefs()
+        preferencesScope.launch(Dispatchers.Main.immediate) {
```
sdsantos: Why do you need to specify again the `Dispatchers.Main.immediate` context in `launch` if it's already set in the `CoroutineScope` constructor?

sdsantos (approved):
All appears to be working fine, and the migration tests I did went smoothly.

app/src/main/java/app/grapheneos/camera/CamConfig.kt
m4pl: It does block. Will be fixed right after [Decompose CamConfig](https://github.com/GrapheneOS/Camera/issues/704), when the screen moves to a [ViewModel](https://github.com/GrapheneOS/Camera/issues/709).

app/src/main/java/app/grapheneos/camera/CamConfig.kt
m4pl: You're right, dropped.

inthewaves (approved):
Looks good, I don't think there's not much else we can do if IOExceptions occur for those calls

PR #719, opened 2026-09-17
Move camera out of MainActivity

m4pl:
Closes #716


app/src/main/java/app/grapheneos/camera/ui/viewfinder/screen/PreviewFrameHolder.kt
```diff
@@ -0,0 +1,188 @@
...
+                else -> lastFrame = previewView.bitmap
+            }
+        }
+    }
+
+    // Starts a copy of the preview for [holdCurrentFrame] to pick up. previewView.bitmap blocks the
+    // caller on a GPU readback for about a tenth of a second, and startCamera() reads it at the
+    // point where it can least afford to block; the same pixels copied asynchronously cost the main
+    // thread nothing, as long as the copy is started early enough.
+    fun prefetch() {
```
sdsantos: Feels a bit weird that one method belongs in the interface, while other methods are just public and the implementation is used directly. I understand that this choice may be because we're refactoring bit-by-bit, specially the `MainActivity`, and that it may not be the final interface. Just wanted to flag it.

sdsantos (changes requested):
Found a small UI bug:
- Zoom slider now always pops up when opening app, switching tabs or switch camera lens

All the rest seems to be working fine.

app/src/main/java/app/grapheneos/camera/data/camera/session/CameraSession.kt
```diff
@@ -586,8 +524,8 @@ internal class CameraSessionImpl @AssistedInject constructor(
             includesVideoCapture = !settings.isQrMode && settings.isVideoMode,
             includesImageCapture = !settings.isQrMode && !settings.requiresVideoModeOnly,
             aspectRatio = settings.aspectRatio,
-            imageCaptureTargetRotation = imageCapture?.targetRotation ?: settings.rotation,
-            previewTargetRotation = preview?.targetRotation ?: settings.rotation,
+            imageCaptureTargetRotation = imageCapture?.targetRotation ?: displayRotation(),
```
RankoR: Rotation is obtained from the default display, not from the current one.

RankoR (approved):

